PR #17, opened 2018-12-14
Add Gradle 5 Info

rushtong:
The render configs and deploy tasks are broken under Gradle 5, so add some clarity around using `gradlew` instead of native `gradle` to run tasks.


rushtong:
You're welcome! I found out the hard way after homebrew upgraded my gradle install to 5 and all of a sudden, nothing worked :-(

PR #22, opened 2018-12-28
ORSP - Project Generation && New Consent Group generation 

vvicario:
The ORSP team would like to re-configure some of their current intake forms for generating new projects, as well as adding intake forms which determine the type of project, and confirm the project's GDPR compliance.

https://broadinstitute.atlassian.net/browse/BTRX-282

The ORSP team would like to re-configure some of their current intake forms for generating new Consent Group and confirm the project's GDPR compliance.

https://broadinstitute.atlassian.net/browse/BTRX-473

---


src/main/webapp/components/Fundings.js
```diff
@@ -0,0 +1,149 @@
...
+            ])
+          ]),
+          div({ className: "col-lg-1 col-md-2 col-sm-2 col-3" }, [
+            Btn({ action: { labelClass: "glyphicon glyphicon-plus", handler: this.addFundings }, disabled: false }),
+          ])
+        ]),
+
+        hr({ className: "fullWidth" }),
+
+        this.props.fundings.map((rd, Index) => {
```
dags: should not use Index as name for, well, index. Use lowercase 'index', 'idx' or similar.

src/main/webapp/components/Fundings.js
```diff
@@ -0,0 +1,149 @@
...
+          return h(Fragment, { key: Index }, [
+
+            div({ className: "row" }, [
+              div({ className: "col-lg-11 col-md-10 col-sm-10 col-9" }, [
+                div({ className: "row" }, [
+                  div({ className: "col-lg-4 col-md-4 col-sm-4 col-12" }, [
+                    InputFieldSelect({
+                      label: "",
+                      id: Index,
```
dags: Index is used as id for all components in div. Should be something like "source" + Index instead. Otherwise source, sponsor and identifier have same id.

src/main/webapp/components/Fundings.js
```diff
@@ -0,0 +1,149 @@
...
+                      onChange: this.handleFundingChange
+                    })
+                  ])
+                ])
+              ]),
+              div({ className: "col-lg-1 col-md-2 col-sm-2 col-3", style: { "paddingTop": "12px" } }, [
+                Btn({ action: { labelClass: "glyphicon glyphicon-remove", handler: (e) => this.removeFundings(Index) }, disabled: !this.state.fundings.length > 1 }),
+              ])
+            ]),
+//            small({ isRendered: this.props.error && Index === 0, className: "errorMessage" }, [this.props.errorMessage])
```
dags: if not used, should remove this comment.

src/main/webapp/components/InputFieldSelect.js
```diff
@@ -0,0 +1,24 @@
...
+import { hh } from 'react-hyperscript-helpers';
+import { InputField } from './InputField';
+import './InputField.css';
+
+export const InputFieldSelect = hh(class InputFieldSelect extends Component {
+
+  render() {
+    return (
+      InputField({ label: this.props.label, moreInfo: this.props.moreInfo, error: this.props.error, errorMessage: this.props.errorMessage }, [
+        <Select
```
dags: Should we use react-hyperscript-helpers here too ?. h(Select, { ....

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -0,0 +1,76 @@
...
+import org.broadinstitute.orsp.ConsentCollectionLink
+import org.broadinstitute.orsp.Issue
+import org.broadinstitute.orsp.IssueType
+
+import javax.ws.rs.core.Response
+
+@Resource(readOnly = false, formats = ['JSON', 'APPLICATION-MULTIPART'])
+class NewConsentGroupController extends AuthenticatedController {
+
+    def fileName = 'Broad_DUL_Draft-Cover_Letter_Form_Fillable.pdf'
```
rushtong: Prefer full object references over dynamic typing when possible.
```suggestion
    private String fileName = 'Broad_DUL_Draft-Cover_Letter_Form_Fillable.pdf'
```

grails-app/domain/org/broadinstitute/orsp/IssueExtraProperty.groovy
```diff
@@ -27,6 +27,8 @@ class IssueExtraProperty {
     public static final String PM = "pm"
     public static final String SOURCE = "source"
     public static final String ACTOR = "actor"
+    public static final String PROJECT_QUESTIONARE = "questions"
+    public static final String PROJECT_TITLE = "pTitle"
```
rushtong: Prefer longer descriptions here. This value gets saved as the name in the issue extra property table so having a better value will make it easier to understand when looking at the database directly.
```suggestion
    public static final String PROJECT_TITLE = "projectTitle"
```

grails-app/domain/org/broadinstitute/orsp/IssueExtraProperty.groovy
```diff
@@ -40,6 +42,10 @@ class IssueExtraProperty {
     public static final String DATA_SHARING_BROAD = "dataSharingBroad"
     public static final String DATA_SHARING_COMMENTS = "dataSharingComments"
     public static final String DATA_SHARING_NIH = "dataSharingNih"
+    public static final String COLLABORATORS = "collaborators"
+    public static final String PROJECT_MANAGER = "pm"
+    public static final String IRB_PROTOCOL_ID = "protocol"
```
rushtong: Note that we already have a `protocol` on line 12. This line isn't necessary and will actually be a conflict.

grails-app/domain/org/broadinstitute/orsp/IssueExtraProperty.groovy
```diff
@@ -63,6 +69,7 @@ class IssueExtraProperty {
     String name
     String value
     String projectKey
+    List<IssueExtraProperty> values
```
rushtong: I'm not a fan of the idea of nesting this object on itself. Can we look into alternatives to this approach?

grails-app/domain/org/broadinstitute/orsp/IssueExtraProperty.groovy
```diff
@@ -40,6 +42,10 @@ class IssueExtraProperty {
     public static final String DATA_SHARING_BROAD = "dataSharingBroad"
     public static final String DATA_SHARING_COMMENTS = "dataSharingComments"
     public static final String DATA_SHARING_NIH = "dataSharingNih"
+    public static final String COLLABORATORS = "collaborators"
```
rushtong: With the addition of collaborators, `PROJECT_MANAGER` becomes unused. This might require a db migration to move all PMs to something else.

grails-app/domain/org/broadinstitute/orsp/IssueExtraProperty.groovy
```diff
@@ -27,6 +27,8 @@ class IssueExtraProperty {
     public static final String PM = "pm"
     public static final String SOURCE = "source"
     public static final String ACTOR = "actor"
+    public static final String PROJECT_QUESTIONARE = "questions"
```
rushtong: Typo:
```suggestion
    public static final String PROJECT_QUESTIONNAIRE = "questions"
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -156,6 +165,52 @@ class IssueService {
         issue
     }
 
+    Issue createIssue(IssueType type, Issue issue) throws DomainException {
```
rushtong: These methods are basically duplicates of `addIssue` and `updateIssue`.

grails-app/views/consentGroup/list.gsp
```diff
@@ -7,6 +7,7 @@
 <body>
 
 <button class="btn btn-default" data-toggle="modal" data-target=".add-new-consent-modal">Add Consent Group</button>
+<a id="" type="button" class="btn btn-default" href="${createLink(controller: 'newConsentGroup', action: 'show', params: [projectKey: issue.projectKey])}" >New Add Consent Group</a>
```
rushtong: Do we need the empty id tag?

package.json
```diff
@@ -46,12 +46,14 @@
     "bootstrap": "4.1.2",
     "lodash-fp": "0.10.4",
     "prop-types": "15.6.2",
-    "react": "16.6.0",
+    "react": "^16.6.3",
```
rushtong: I have a question about using `^`. I have preferred to tie the versions exactly to avoid problems with library updates causing conflicts. It happened a lot in the angular version of consent-ui. Is there a reason to use the `compatible-with` syntax instead of the exact version?

src/main/groovy/org/broadinstitute/orsp/FundingFields.groovy
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.orsp
+
+enum FundingFields {
```
rushtong: Is this used?

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -14,7 +14,9 @@ class UrlMappings {
         "/api/report/funding"(controller: 'report', action: "funding")
         "/"(view:"/index")
         '/api/swagger/**'(controller: 'api', action: 'swagger')
-
+        '/api/project'(resource: 'project')
+        '/api/files-helper/attach-document'(controller: 'fileHelper', action: 'attachDocument', method: 'POST')
+        '/api/newConsentGroup'(resource: 'newConsentGroup')
```
rushtong: On lines 18 and 19, we have two different styles of url here. Grails, by default, uses camel casing based on controller names (i.e. `ConsentGroupController` -> `/consentGroup`) which is not unreasonable. I actually like `kebab-case` urls so if we can try to be consisnt with that as we move forward, that would be my preference.

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -60,39 +66,116 @@ class Issue {
...
+        getExtraProperties().findAll { it.name == IssueExtraProperty.PM }.collect { it.value }
+    }
+
+    transient Collection<String> getAffiliations() {
+        getExtraProperties().findAll { it.name == IssueExtraProperty.AFFILIATIONS }.collect { it.value }
+    }
+
+    transient Collection<String> getNotResearch() {
+        getExtraProperties().findAll { it.name == IssueExtraProperty.NOT_RESEARCH }.collect { it.value }
+    }
```
rushtong: I would prefer to revert this formatting change. For large, repetitive blocks like this, I like the condensed format better.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -45,14 +46,21 @@ class IssueService {
             IssueExtraProperty.RESPONSIBLE,
             IssueExtraProperty.REVIEW_CATEGORY,
             IssueExtraProperty.SOURCE,
-            IssueExtraProperty.SUBMISSION_TYPE]
+            IssueExtraProperty.SUBMISSION_TYPE,
+            IssueExtraProperty.PROJECT_TITLE,
```
rushtong: Some history here. For the thing we describe as title in the UI, we actually use the `summary` field on Issue. Issue, originally, corresponded to a Jira Issue since we used Jira as a back end for this app for the first couple years of its life. We then moved to its own database when it outgrew Jira. In Jira, there are a small number of primary fields on the Issue object, and then all sorts of custom fields can be defined that are saved in a separate table. The distinction between where we store things is a bit arbitrary, but in general, really important things are on the `Issue` object itself, while less important things were stored as custom fields. Custom fields became `IssueExtraProperty`.

The point here is that I think we should keep `title`, but move it to the issue directly via a liquibase change. Also, the content of that field should start with the value currently stored in `summary`. We could then use both fields more appropriately in the UI. 

However, this change seems like it should be in a separate PR for clarity.

webpack.config.js
```diff
@@ -6,7 +6,10 @@ module.exports = {
...
   entry: {
     collectionLinks: './src/main/webapp/collectionLinks/index.js',
-    search: './src/main/webapp/search/index.js'
+    search: './src/main/webapp/search/index.js',
+    pages: './src/main/webapp/pages/index.js',
+    consentGroup: './src/main/webapp/consentGroup/index.js'
+
   },
   output: {
     path: path.join(__dirname, 'grails-app/assets/javascripts'),
```
rushtong: One thing I meant to do was to figure out how to update the build process so we don't have to check in the compiled versions of these `index.js` files. The fact that there is a prod/dev version of these can lead to problems if we check in the wrong version. Not for this PR, but we should add a story that does the following:

* Update the build task (war) to run the prod webpack command before building.
* Move all compiled js references to point to a new directory, something like `grails-app/assets/javascripts/compiled`
* Change this output path to match
* Set that directory to be ignored in `.gitignore`

I think that would be enough to get us away from checking in the compiled js code.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
wloforteBroad: @rushtong We agree, maybe you can create a story for this (If you can) so we can tackle this change in the near future.

webpack.config.js
wloforteBroad: @rushtong I can take care of that and put it in our "For Refinement" backlog.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
LeoForconesi: The idea is to remove old duplicated code as we move forward in the react migration functionality,  we kept this temporally in order to maintain stability on the system with previous info. 

We still need to work on some methods using the new logic for new Projects and new Consent group creation, to finally clean up. The plan is to create a Jira story in order to accomplish this, if it's ok for you.

rushtong: That sounds like a good plan, thanks for the explanation! 👍

rushtong: Not required, but might be nice to add `@Deprecated` to the things we plan on removing so we can keep track better.

vvicario:
@rushtong we worked on your feedback. Please let us know if you have any other comment. Thanks!!

rushtong:
@vvicario FYI - I moved this PR to go against `feature`. What I'd like to do is use `feature` as the main working branch for all of these commits until the new workflow is complete. Once that's done, we can merge `feature` into `master`.

vvicario:
@rushtong sounds good, Thank you!

grails-app/views/newProject/index.gsp
```diff
@@ -0,0 +1,38 @@
+<%@ page import="org.broadinstitute.orsp.IssueStatus" %>
+<%@ page import="org.broadinstitute.orsp.IssueType" %>
+<%@ page import="org.broadinstitute.orsp.PreferredIrb" %>
```
rushtong: pretty sure these aren't used

PR #24, opened 2019-01-09
[BTRX-489] Show errors form

tadeork:
## Addresses
Jira story [BTRX-489](https://broadinstitute.atlassian.net/browse/BTRX-489)
## Changes
Show errors when something fails in the server.
## Testing

---

### DIFF
--- a/src/main/webapp/consentGroup/NewConsentGroup.js
+++ b/src/main/webapp/consentGroup/NewConsentGroup.js
@@ -16,6 +16,7 @@ class NewConsentGroup extends Component {
       showErrorStep3: false,
       generalError: false,
       formSubmitted: false,
+      submitError: false,
       determination: {
         projectType: 900,
         questions: [],
@@ -58,6 +59,8 @@ class NewConsentGroup extends Component {
     this.downloadFillablePDF = this.downloadFillablePDF.bind(this);
     this.submitNewConsentGroup = this.submitNewConsentGroup.bind(this);
     this.uploadFiles = this.uploadFiles.bind(this);
+    this.changeStateSubmitButton = this.changeStateSubmitButton.bind(this);
+    this.toggleSubmitError = this.toggleSubmitError.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -617,12 +638,60 @@ class NewConsentGroup extends Component {
...
+          title: "Data Sharing",
+          currentStep: currentStep,
+          user: this.props.user,
+          searchUsersURL: this.props.searchUsersURL,
+          updateForm: this.updateStep5FormData,
+          errors: this.state.errors,
+          removeErrorMessage: this.removeErrorMessage,
+          generalError: this.state.generalError,
+          submitError: this.state.submitError
+        }),
```
rushtong: Thank you for unwrapping these - I think this format is more legible.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -58,6 +59,8 @@ class NewConsentGroup extends Component {
     this.downloadFillablePDF = this.downloadFillablePDF.bind(this);
     this.submitNewConsentGroup = this.submitNewConsentGroup.bind(this);
     this.uploadFiles = this.uploadFiles.bind(this);
+    this.changeStateSubmitButton = this.changeStateSubmitButton.bind(this);
+    this.toggleSubmitError = this.toggleSubmitError.bind(this);
```
rushtong: Not for this PR - but I do have a question. I've found that it is necessary to bind the functions this way but I do find it verbose and unintuitive boilerplate. I'd love to be able to get rid of this if there's a better way to do this.

src/main/webapp/project/NewProjectDocuments.js
```diff
@@ -46,11 +46,23 @@ export const NewProjectDocuments = hh(class NewProjectDocuments extends Componen
...
+    if (this.props.submitError) {
+      errorText = 'Something went wrong in the server. Please try again later.';
+    }
+
     return (
 
       WizardStep({
         title: this.props.title, step: 2, currentStep: this.props.currentStep,
-        errorMessage: !this.props.generalError ? 'Please upload all required documents' : 'Please check previous steps',
+        errorMessage: errorText,
```
rushtong: Nice refactoring.

rushtong:
Looks good pending Diego's approval 👍

src/main/webapp/consentGroup/NewConsentGroup.js
tadeork: I'll check if there's a better way to do this binding.

src/main/webapp/consentGroup/NewConsentGroup.js
tadeork: (y) I think the same

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -81,17 +84,22 @@ class NewConsentGroup extends Component {
 
   submitNewConsentGroup = () => {
 
+    this.setState(prev => {
```
dags: for future use : we should plain _**this.setState({ submitError: false });**_ format when possible. No need to use "prev" approach here.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -101,19 +109,32 @@ class NewConsentGroup extends Component {
     }
   };
 
+  changeStateSubmitButton = () => {
```
dags: please change to changeSubmitState ...

PR #25, opened 2019-01-09
[BTRX-522] Endpoints Approve Reject documents

tadeork:
## Changes
Added two new endpoints: approve and reject document. It was added in `FileHelperController.groove` because there is the handling of file documents. Also changed the way the file is uploaded to add a new field "Pending". Changed the http status code. New changelog to add the new field in `storage_document` and also a default value to the older documents.
In the view is a small change to fix the file key spelling.
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/FileHelperController.groovy
```diff
@@ -15,17 +25,59 @@ class FileHelperController extends AuthenticatedController{
             files.forEach {
                 names.push(it.name)
                 if (!files.empty) {
-                    storageProviderService.saveMultipartFile(
-                            (String) params.displayName,
-                            (String) params.userName,
-                            (String) issue.projectKey,
-                            (String) it.name,
-                            (MultipartFile) it)
+                    def document = new StorageDocument(
```
rushtong:
```suggestion
                    StorageDocument document = new StorageDocument(
```

grails-app/controllers/org/broadinstitute/orsp/api/FileHelperController.groovy
```diff
@@ -1,10 +1,20 @@
...
+import org.broadinstitute.orsp.StorageDocument
 import org.springframework.web.multipart.MultipartFile
 
+import java.text.SimpleDateFormat
+
+@Resource(readOnly = false, formats = ['JSON', 'APPLICATION-MULTIPART'])
 class FileHelperController extends AuthenticatedController{
+    private static final String APPROVED = "Approved"
+    private static final String REJECTED = "Rejected"
+    private static final String PENDING = "Pending"
```
rushtong: I think these strings should all be elements of an `Enum` instead of individual values. You can also add the `Legacy` one as an enum option so old documents can be displayed correctly, if necessary.

PR #34, opened 2019-01-15
Documents view for Project and documents view for Consent Group

tadeork:
## Addresses
Jira stories [BTRX-519](https://broadinstitute.atlassian.net/browse/BTRX-519) [BTRX-520](https://broadinstitute.atlassian.net/browse/BTRX-520) [BTRX-521](https://broadinstitute.atlassian.net/browse/BTRX-521)
## Changes
New views for Documents tab in Consent Group and Project. It checks if is the user is Admin; admin can change status from Pending to Approved or Rejected.
## Testing

---


grails-app/domain/org/broadinstitute/orsp/StorageDocument.groovy
```diff
@@ -31,6 +32,10 @@ class StorageDocument {
         status nullable: false
     }
 
+    static mapping = {
+        version column: '`version`'
```
rushtong: Can we use a different name here, something like `document_version`? Grails/GORM auto-adds a `version` column to all mapped entities and I don't want to get confused between what it means here as compared to other entities.

grails-app/views/consentGroupDocuments/_index.gsp
```diff
@@ -0,0 +1,24 @@
+<g:javascript>
+    const user = {
+        <g:if test="${session?.user}">
+            displayName: "${session.user.displayName}",
+            email: "${session.user.emailAddress}",
+            userName: "${session.user.userName}"
+        </g:if>
+    };
```
rushtong: See https://github.com/broadinstitute/orsp-pub/pull/33#discussion_r248318817

I think this approach is not the best one to use. The point of `const component` is that it is just a holder for constant values. We should instead pass in a URL that the component can call to get the current user. In the long term, these `const component` values will be rolled up into a single configuration component that the app has access to.

grails-app/views/consentGroupDocuments/_index.gsp
```diff
@@ -0,0 +1,24 @@
...
+    const component = {
+        user: user,
+        issue: '${issue.requestDate}',
+        projectKey:'${issue.projectKey}',
```
rushtong: Also see that same comment. I'd like to try and abstract out anything that is a variable to a service call to get the variable value at the time the component needs to load it. 

Also, `issue: '${issue.requestDate}',` is confusing. Is it `issue` or is it `requestDate` that you want here?

grails-app/views/irb/show.gsp
```diff
@@ -66,13 +67,22 @@
...
+    <div id="documentsNew">
+        <g:render template="/projectDocument/index" model="[
                 controller        : issue.controller,
                 issue             : issue,
-                attachmentTypes   : attachmentTypes,
-                storageDocuments  : storageDocuments
-        ]}"/>
+                attachmentTypes   : attachmentTypes
+                ]"/>
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/react-bootstrap-table/4.3.1/react-bootstrap-table-all.min.css"/>
```
rushtong: I've seen this in the other PR too. It doesn't make sense here - it should either be in the component or a global style ref.

grails-app/views/ne/show.gsp
```diff
@@ -66,15 +67,24 @@
...
         ]}"/>
     </div>
 
+    <div id="documentsNew">
+        <g:render template="/projectDocument/index" model="[
+            controller        : issue.controller,
+            issue             : issue,
+            attachmentTypes   : attachmentTypes
+          ]"/>
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/react-bootstrap-table/4.3.1/react-bootstrap-table-all.min.css"/>
```
rushtong: See the `show` page comment.

grails-app/views/nhsr/show.gsp
```diff
@@ -66,15 +67,23 @@
...
+          storageDocuments  : storageDocuments
         ]}"/>
     </div>
 
+    <div id="documentsNew">
+        <g:render template="/projectDocument/index" model="[
+                controller        : issue.controller,
+                issue             : issue
+                ]"/>
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/react-bootstrap-table/4.3.1/react-bootstrap-table-all.min.css"/>
```
rushtong: More :-)

grails-app/views/projectDocument/_index.gsp
```diff
@@ -0,0 +1,26 @@
+<g:javascript>
+    const user = {
+        <g:if test="${session?.user}">
+            displayName: "${session.user.displayName}",
+            email: "${session.user.emailAddress}",
+            userName: "${session.user.userName}"
+        </g:if>
+    };
```
rushtong: See previous user service comment.

grails-app/views/projectDocument/_index.gsp
```diff
@@ -0,0 +1,26 @@
...
+      user: user,
+      projectKey:'${issue.projectKey}',
+      attachedDocumentsUrl: "${createLink(uri: '/api/files-helper/attached-documents', method: 'GET')}",
+      attachDocumentsUrl: "${createLink(uri: '/api/files-helper/attach-document', method: 'POST')}",
+      issue: '${issue.requestDate}',
+      projectUrl: "${createLink(controller: 'project', action: 'getProject')}",
+      roles: '${session.roles}'.replace(/\s+/g, '').replace("[", "").replace("]", "").split(","),
+      rejectDocumentUrl: "${createLink(uri: '/api/files-helper/reject-document', 'PUT')}",
+      approveDocumentUrl: "${createLink(uri: '/api/files-helper/approve-document', method: 'PUT')}",
+      isAdmin: JSON.parse("${session.isAdmin}"),
```
rushtong: This could be on the user instead of a session call.

grails-app/views/projectDocument/_index.gsp
```diff
@@ -0,0 +1,26 @@
...
+            email: "${session.user.emailAddress}",
+            userName: "${session.user.userName}"
+        </g:if>
+    };
+    const component = {
+      user: user,
+      projectKey:'${issue.projectKey}',
+      attachedDocumentsUrl: "${createLink(uri: '/api/files-helper/attached-documents', method: 'GET')}",
+      attachDocumentsUrl: "${createLink(uri: '/api/files-helper/attach-document', method: 'POST')}",
+      issue: '${issue.requestDate}',
```
rushtong: This could be derived from a call to get project by project key, which is above.

grails-app/views/projectDocument/_index.gsp
```diff
@@ -0,0 +1,26 @@
...
+      projectUrl: "${createLink(controller: 'project', action: 'getProject')}",
+      roles: '${session.roles}'.replace(/\s+/g, '').replace("[", "").replace("]", "").split(","),
```
rushtong: We should not be referencing session in these GSP pages. All variable content should be queried via a service call.

src/main/webapp/components/Documents.js
```diff
@@ -0,0 +1,52 @@
+import { Component } from 'react';
+import { input, hh, h, div, p, hr, small, button } from 'react-hyperscript-helpers';
+import { Table } from './Table';
+import { Panel } from './Panel';
+
+ const headers =
+  [
```
rushtong: Nit: the spacing is off in this file. If you reformat the lines, you'll see a lot of whitespace changes.

src/main/webapp/components/Table.css
```diff
@@ -0,0 +1,22 @@
+.tableContainer .table  {
+    margin-bottom: 0;
+}
+
+ .tableContainer .table-search-form {
```
rushtong: Nit: spacing - same problem as `Documents.js`

src/main/webapp/components/Table.js
```diff
@@ -0,0 +1,90 @@
+import { Component } from 'react';
+import { a, hh, h, div, p, hr, small } from 'react-hyperscript-helpers';
+import React from 'react';
+import {BootstrapTable, TableHeaderColumn} from 'react-bootstrap-table';
+import { DropdownButton, MenuItem, ButtonToolbar } from 'react-bootstrap';
+import './Table.css';
+
+ export const Table = hh(class Table extends Component {
```
rushtong: Nit: spacing - same problem as `Documents.js`

src/main/webapp/components/Table.js
```diff
@@ -0,0 +1,90 @@
+import { Component } from 'react';
+import { a, hh, h, div, p, hr, small } from 'react-hyperscript-helpers';
+import React from 'react';
+import {BootstrapTable, TableHeaderColumn} from 'react-bootstrap-table';
```
rushtong: nit: spacing - rest of the imports have spaces after/before the braces. Let's try to be consistent.
```suggestion
import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';
```

src/main/webapp/components/Table.js
```diff
@@ -0,0 +1,90 @@
...
+                                        dataFormat={this.formatStatusColumn}
+                                        dataSort={true}>{header.name}</TableHeaderColumn>
+            } else if (header.value === 'fileName') {
+              return <TableHeaderColumn key={header.name}
+                                        dataField={header.value}
+                                        dataFormat={this.formaUrlDocument}
+                                        dataSort={true}>{header.name}</TableHeaderColumn>
+            } else {
+              return <TableHeaderColumn key={header.name} dataField={header.value} dataSort={ true }>{header.name}</TableHeaderColumn>
+            }
```
rushtong: Ton of duplicated code here. To help sort this out, I would pre-calculate the things that are different between the three `TableHeaderColumn` components then use a single `TableHeaderColumn` with the calculated values. I think the things that can be pre-calculated are:
* isKey: `isKey = (index === 0)`
* dataFormat: `dataFormat = (header.value === 'status') ? this.formatStatusColumn : this.formaUrlDocument` and maybe add a third condition to set `dataFormat = undefined`

Then you can have a single return:
```
return <TableHeaderColumn 
                                        key={header.name}
                                        isKey={isKey}
                                        dataField={header.value}
                                        dataFormat={dataFormat}
                                        dataSort={true}>{header.name}</TableHeaderColumn>
```
Caveat ... untested code!

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -0,0 +1,127 @@
...
+    DocumentHandler.attachedDocuments(this.props.attachmentsUrl, this.props.projectKey).then(resp => {
+      this.setKeyDocuments(JSON.parse(resp.data.documents));
+    }).catch(error => {
+      console.error(error);
+    });
+  };
+
+   setKeyDocuments = (documentsCollection) => {
+    const keyDocuments = [];
+    const additionalDocuments = [];
```
rushtong: These are constants that you are modifying. Should they be `let`s instead?

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -0,0 +1,127 @@
...
+   handleChangeStatus = (uuid, status) => {
+    const formerStateKeyDoc = this.state.keyDocuments.slice();
+    formerStateKeyDoc.forEach(doc => {
+      if (uuid === doc.uuid) {
+        doc.status = status;
+      }
+    });
+    this.setState({ keyDocuments: formerStateKeyDoc });
+
+     const formerAdditionalDoc = this.state.additionalDocuments.slice();
```
rushtong: nit: indentation here and throughout this file.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -0,0 +1,127 @@
...
+        handleOkAction: this.handleAction,
+        title: this.state.action + ' Confirmation',
+        bodyText: 'Are you sure yo want to ' + this.state.action + ' this document?',
+        actionLabel: 'Yes'
+      }, []),
+      Documents({
+        keyDocuments: this.state.keyDocuments,
+        additionalDocuments: this.state.additionalDocuments,
+        handleDialogConfirm: this.handleDialog,
+        isAdmin: this.props.isAdmin,
```
rushtong: It seems like this should be closer to the `Documents` component and is really just a modifier on the user. Is there any reason we can't pass the user down to this component instead? That way we can let `Documents` make the decision about what to show based on anything it cares to.

src/main/webapp/consentGroupDocuments/index.js
```diff
@@ -0,0 +1,18 @@
...
+ ReactDOM.render(
+  <ConsentGroupDocuments
+    attachmentsUrl = {component.attachmentsUrl}
+    serverURL = {component.serverURL}
+    user = {component.user}
+    attachDocumentsUrl = {component.attachDocumentsUrl}
+    projectKey = {component.projectKey}
+    approveDocumentUrl = {component.approveDocumentUrl}
+    rejectDocumentUrl = {component.rejectDocumentUrl}
+    isAdmin = {component.isAdmin}
```
rushtong: See previous comments about this - I think this should be part of the `user` object and not necessarily a separate prop.

src/main/webapp/projectDocument/ProjectDocument.js
```diff
@@ -0,0 +1,127 @@
+import { Component, Fragment } from 'react';
+import { Documents } from '../components/Documents'
+import { DocumentHandler } from "../util/ajax";
+import { ProjectKeyDocuments } from '../util/KeyDocuments';
+import { ConfirmationDialog } from "../components/ConfirmationDialog";
+import { h } from 'react-hyperscript-helpers';
+
+ class ProjectDocument extends Component {
```
rushtong: Nit: indentation is off in this file.

src/main/webapp/projectDocument/ProjectDocument.js
```diff
@@ -0,0 +1,127 @@
...
+
+   componentDidMount() {
+    this.getAttachedDocuments();
+  }
+
+   getAttachedDocuments = () => {
+    DocumentHandler.attachedDocuments(this.props.attachedDocumentsUrl, this.props.projectKey).then(resp => {
+      this.setKeyDocuments(JSON.parse(resp.data.documents));
+    }).catch(error => {
+      console.error(error);
```
rushtong: Note: We really need a way to handle JS errors for the user more gracefully than just console-ing them.

src/main/webapp/projectDocument/ProjectDocument.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+    this.closeModal();
+  };
+
+   closeModal = () => {
+    this.setState({showDialog: !this.state.showDialog});
+  };
+
+   render() {
+    console.log("IS ADMINNN:  - ", this.props.isAdmin);
```
rushtong: remove?

src/main/webapp/projectDocument/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+
+ ReactDOM.render(
+    <ProjectDocument
+        projectKey = {component.projectKey}
+        attachDocumentsUrl = {component.attachDocumentsUrl}
+        attachedDocumentsUrl = {component.attachedDocumentsUrl}
+        approveDocumentUrl = {component.approveDocumentUrl}
+        rejectDocumentUrl = {component.rejectDocumentUrl}
+        isAdmin = {component.isAdmin}
```
rushtong: See previous comments about this.

src/main/webapp/projectDocument/index.js
```diff
@@ -0,0 +1,20 @@
...
+        component = {component}
+        downloadDocumentUrl = {component.downloadDocumentUrl}
+        user = {component.user}
```
rushtong: See previous comments about this. We should probably be fetching the user before this is rendered and passing that in.

src/main/webapp/util/KeyDocuments.js
```diff
@@ -0,0 +1,14 @@
...
+  'NE Approval',
+  'NE Application',
+  'Consent Document'
+];
+
+ export const ConsentGroupKeyDocuments = [
+  'Approval Memo',
+  'Consent Document',
+  'Data Use Letter'
+];
```
rushtong: Nit: newline

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
tadeork: I think this is OK bcs these are arrays and the values are worked with `push()` - `pop()`. Unless we need to redefine the type of the variable then `let` should be the way, if not `const` is like a standar.

PR #36, opened 2019-01-17
[risk=no] Update the get session user call

rushtong:
Update the `getSessionUser` call so it includes user and role information. Does not remove current calls to `isAdmin` so those should be dealt with separately.

Also - checked in the prod-compiled js files as a result of running the `webpackProd` gradle task. That accounts for the monstrous number of line changes.

End result is that this endpoint now returns something like this:
```
{
    "displayName": "Gregory Rushton",
    "emailAddress": "grushton@broadinstitute.org",
    "userName": "grushton",
    "isAdmin": true,
    "isORSP": true,
    "isComplianceOffice": true
}
```


grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
```diff
@@ -347,12 +347,24 @@ class AuthenticatedController implements Interceptor, UserInfo {
...
+    def getSessionUser() {
+        User user = getUser()
+        render([
+            'displayName': user.displayName,
+            'emailAddress': user.emailAddress,
+            'userName': user.userName,
+            'isAdmin': isAdmin(),
+            'isORSP': isORSP(),
+            'isComplianceOffice': isComplianceOffice(),
+        ] as JSON)
```
rushtong: The big changes here are that we return a single object instead of an array and dump in some additional, relevant properties.

Side note: I first tried adding transients to `User`, but `JSON` ignores transient properties. This turned out to be simpler since I can customize the exact properties to return. We should probably not return the db id, role ids, etc.

LeoForconesi: I like this way better than how it was done before. And its reutilizable. :+1:

vvicario: Nice. I'll update our PR with this approach. Thanks!

PR #38, opened 2019-01-18
BTRX - 528 - Project Approve

LeoForconesi:
## Addresses
[BTRX-528](https://broadinstitute.atlassian.net/browse/BTRX-528): Implement General Approval of Project
## Changes
- Add a common method to verify if a project (issue) will be approved. There are two conditions for this to happen. 
    **1 -** Project general data review must be approved; 
    **2 -** Every Project documents must be approved
If both conditions are fulfilled, then the project will be Approved
---


grails-app/controllers/org/broadinstitute/orsp/api/FileHelperController.groovy
```diff
@@ -66,6 +67,8 @@ class FileHelperController extends AuthenticatedController{
         try {
             if (document != null) {
                 document.setStatus(DocumentStatus.APPROVED.status)
+                Issue issue = queryService.findByKey(document.projectKey)
+                issueService.projectApproval(issue)
```
dags: please group document behavior ...

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +211,35 @@ class IssueService {
...
         updatedIssue.save(flush:true)
         updatedIssue
     }
 
+    void projectApproval(Issue issue) {
+        Boolean projectGeneralDataApproved = false
+        Boolean projectDocumentsApproved = false
+
+        if (issue != null && !issue.getAttachments().isEmpty()) {
+            issue.getNonEmptyExtraProperties()?.each{
```
dags: Please try to use filter or find here instead of each

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +211,35 @@ class IssueService {
...
+                if (it.name == IssueExtraProperty.PROJECT_REVIEW_APPROVED) {
+                    projectGeneralDataApproved = it.value
+                }
+            }
+            projectDocumentsApproved = issue.getAttachments().collect {
```
dags: here could use find too

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
LeoForconesi: Done!

grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
```diff
@@ -40,8 +41,17 @@ class ProjectController extends AuthenticatedController {
         String projectKey = params.id
         Gson gson = new Gson()
         Object input = gson.fromJson(gson.toJson(request.JSON), Object.class)
-        Issue updatedIssue = issueService.modifyExtraProperties(input, projectKey)
-        render([message: updatedIssue] as JSON)
+        try {
+            Issue updatedIssue = issueService.modifyExtraProperties(input, projectKey)
+            if (input.get(IssueExtraProperty.PROJECT_REVIEW_APPROVED)) {
```
rushtong: I've tried to keep all property getters in Issue - I think this could be there.
In Issue (change to a boolean if that's what this is):
```
transient String getProjectReviewApproved() { getExtraProperties().find { it.name == IssueExtraProperty.PROJECT_REVIEW_APPROVED }?.value }
```
Then here:
```suggestion
            if (issue.getProjectReviewApproved()) {
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +211,34 @@ class IssueService {
...
         updatedIssue.save(flush:true)
         updatedIssue
     }
 
+    void projectApproval(Issue issue) {
+        Boolean projectGeneralDataApproved = false
+        Boolean projectDocumentsApproved = false
+
+        if (issue != null && !issue.getAttachments().isEmpty()) {
+            projectGeneralDataApproved = issue.getNonEmptyExtraProperties().find {
```
rushtong: This could be simplified by my suggestion above:
```suggestion
            projectGeneralDataApproved = issue.getProjectReviewApproved()
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +211,34 @@ class IssueService {
...
+                it.name == IssueExtraProperty.PROJECT_REVIEW_APPROVED
+            }?.value
+
+            projectDocumentsApproved = issue.getAttachments().find {
+                it.status != IssueStatus.Approved.getName()
+            } != null
```
rushtong: Just want to understand this logic better - maybe a comment explaining what this is doing would be nice. In this code, we look for any attachment that **IS NOT** approved .... and if we find any, we are setting `projectDocumentsApproved` to true. 

That seems counter-intuitive to me. I would expect, just looking at variable names, that you would set the value to true if and only if **ALL** attachments were approved.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +211,34 @@ class IssueService {
...
+        }
+
+        if (projectGeneralDataApproved == true && projectDocumentsApproved == true ) {
```
rushtong: Can be simplified ... both are booleans:
```suggestion
        if (projectGeneralDataApproved && projectDocumentsApproved) {
```

grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
LeoForconesi: Makes sense to have a "Project Review Approved" getter in Issue, I added as you suggested. Eitherway the idea here was to be able to call `modifyExtraProperties` with a generic input object. Since `input` can have any extra property we only call `updateProjectApproval` if the current input is intended to change this state.

If we receive an input without "PROJECT_REVIEW_APPROVED" constant, but the issue/project (`updatedIssue` in this case) already has this property, it will verify again all approvals, which won't break anything... anyhow will take some extra db calls.

Please check again the change I made. This check is now performed in `modifyExtraProperties` as you can see [here](https://github.com/broadinstitute/orsp-pub/blob/BTRX-528-generalApprove/grails-app/services/org/broadinstitute/orsp/IssueService.groovy#L204)

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
LeoForconesi: Please check again if the new changes resolve your comment.

LeoForconesi:
@rushtong When you have some time please review these new changes. I'll be pending to your replies

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -205,4 +207,14 @@ class Issue {
         extraproperties
     }
 
+    /** AttachmentApproved verifies if there's any attachment status different to 'Approved'
+     *  if there is any, it will return a value different to null, meaning it still has attachments unReviewed
+     *
+     * @return Boolean indicating if all attachments have 'Approved' status
+     */
```
rushtong: Nit: spacing and language: 
```suggestion
    /** 
     * Determines if there's any status on attachments other than 'Approved', meaning 
     * the issue still has unapproved or unreviewed attachments.
     *
     * @return True if all attachments have the 'Approved' status, false otherwise
     */
```

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -205,4 +207,14 @@ class Issue {
...
+    transient Boolean attachmentApproved(){
```
rushtong: Nit: naming and spacing:
```suggestion
    transient Boolean attachmentsApproved() {
```

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -205,4 +207,14 @@ class Issue {
...
+        getAttachments()?.find {
+            it.status != IssueStatus.Approved.getName()
+        } == null
```
rushtong: Nit: Very minor suggestion. You could also use collect and query the emptiness of the collection. I usually like that approach over checking against null. Definitely not required.
```
        getAttachments()?.collect {
            it.status != IssueStatus.Approved.getName()
        }?.isEmpty()
```
Alternatively, you could use the spread operator over status in combination with `any` which is nice shorthand:
```
getAttachments()?.status?.any { 
    !it.equals(IssueStatus.Approved.getName()) 
}
```
This is an area where Groovy shines and comes closest to a scala-like functional style.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -200,6 +200,10 @@ class IssueService {
         if (!extraPropertiesList.isEmpty()) {
             saveExtraProperties(issue, extraPropertiesList)
         }
+        issue.extraProperties.addAll(extraPropertiesList)
+        if (extraPropertiesList.find {it.name == IssueExtraProperty.PROJECT_REVIEW_APPROVED} != null ) {
```
rushtong: Nit: In groovy, non-null and non-empty are both truthy, so I don't think you need the `!= null` here:
```suggestion
        if (extraPropertiesList.find {it.name == IssueExtraProperty.PROJECT_REVIEW_APPROVED}) {
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -210,12 +214,25 @@ class IssueService {
...
             }
         }
         updatedIssue.save(flush:true)
         updatedIssue
     }
 
+    /** updateProjectApproval
+     *  This method will check if a given issue has its general data and each of its attachment Approved,
+     *  if this conditions are met, then we set its general issue status to 'Approved'
+     * */
```
rushtong: Nit: javadoc:
```suggestion
    /**
     * Check that an issue has its general data and all of its attachments are in 'Approved' status.
     * If all conditions are met, then we set its general status to 'Approved'
     */
```

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
LeoForconesi: Good to know! I used the second suggestion with the spread operator.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
LeoForconesi: Got it!

PR #41, opened 2019-01-22
Reject Project & Consent Group

vvicario:
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -88,4 +88,22 @@ class NewConsentGroupController extends AuthenticatedController {
...
+    def delete() {
+        Issue issue = queryService.findByKey(params.consentKey)
+        if(issue != null) {
+            Collection<StorageDocument> documents = queryService.getDocumentsForProject(params.consentKey)
+            documents?.each {
+                storageProviderService.removeStorageDocument(it, getUser()?.displayName)
+            }
+            def links = ConsentCollectionLink.findAllByProjectKey(params.consentKey)
+            if(links != null) persistenceService.deleteCollectionLinks(links)
+            issueService.deleteIssue(params.consentKey)
```
rushtong: Nit: There's a good amount of business logic here. In general, we should keep controllers thin and move business logic to a service class instead. I think all of this logic could be in `IssueService` instead.

grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
```diff
@@ -61,6 +63,23 @@ class ProjectController extends AuthenticatedController {
...
+    def delete() {
+        Issue issue = queryService.findByKey(params.projectKey)
+        if(issue != null) {
+            Collection<StorageDocument> documents = queryService.getDocumentsForProject(params.projectKey)
+            documents?.each {
+                storageProviderService.removeStorageDocument(it, getUser()?.displayName)
+            }
+            def links = ConsentCollectionLink.findAllByProjectKey(params.projectKey)
+            if(links != null) persistenceService.deleteCollectionLinks(links)
+            issueService.deleteIssue(params.projectKey)
```
rushtong: And ... see my comment above - move to `IssueService`.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -237,6 +237,21 @@ class IssueService {
...
+            def fundingList = queryService.findFundingsByProject(issue.projectKey)
+            fundingList?.each {
+                it.delete()
+            }
+            Collection<IssueExtraProperty> issueExtraProperties = issue.getExtraProperties()
+            issueExtraProperties?.each {
+                it.delete()
+            }
+        }
+        issue.delete(flush: true)
```
rushtong: See https://github.com/broadinstitute/orsp/blob/develop/sql/utility/delete_issue.sql for a full list of things to look at when deleting an issue. And ... thank you for adding this! I run that delete script every other week or so.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -201,8 +202,8 @@ class ConsentGroupReview extends Component {
   }
 
   approveRevision = (e) => () => {
-    this.setState({ disableApproveButton: true })
-    const data = { projectReviewApproved: true }
+    this.setState({ disableApproveButton: true });
+    const data = { projectReviewApproved: true };
```
rushtong: Nit: Throughout, no need for the semi-colons.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -175,7 +181,9 @@ class ProjectReview extends Component {
   }
 
   isAdmin() {
-    return this.props.isAdmin === "true";
+    User.getUserSession(this.props.sessionUserUrl).then(resp => {
+      this.setState({isAdmin: resp.data.isAdmin});
+    });
```
rushtong: We should consider caching this at the page level. Every time `isAdmin` is called, we make a rest call, which is probably unnecessary.

rushtong: Also, consider moving admin to props. A user's admin status doesn't really change so it doesn't need to be in state. Try to use state only for things that change, and compute as much as possible.

build.gradle
```diff
@@ -119,6 +119,7 @@ dependencies {
     compile group: 'org.codehaus.groovy.modules.http-builder', name: 'http-builder', version: '0.7.1'
     compile group: 'io.github.http-builder-ng', name: 'http-builder-ng-okhttp', version: '1.0.3'
     compile group: 'org.apache.httpcomponents', name: 'httpmime', version: '4.5.6'
+    compile 'org.grails.plugins:gorm-logical-delete:2.0.0.M2'
```
rushtong: Once this is merged, there will need to be a lot of updates to the various sql-based queries in `QueryService` and possibly elsewhere. The plugin auto-magically adds a where clause on the deleted field to GORM queries. Raw sql does not do that and we don't do a lot of GORM querying (have been migrating away from that over time).

src/main/webapp/projectReview/ProjectReview.js
vvicario: Greg I reverted my changes, If you agree we will refactor user role code related in US https://broadinstitute.atlassian.net/browse/BTRX-533. 
I finish with logical delete, I'll let you know as soon as walter finish with testing.

build.gradle
vvicario: Greg if you want I can do the  sql-based queries  updates in this branch

rushtong: Not necessary for this branch. Updating the sql queries will only affect issues going through the old flow, not the new, so it makes sense to me that this be a separate effort.

vvicario: ok. Thanks!

vvicario:
@rushtong Please let me know if you have any other comments. Walter finished with testing

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -237,6 +237,13 @@ class IssueService {
         newIssue
     }
 
+    void deleteIssue(String projectKey) {
+        Issue issue = queryService.findByKey(projectKey)
+        if(issue != null) {
```
rushtong: nit: spacing
```suggestion
        if (issue != null) {
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -237,6 +237,13 @@ class IssueService {
...
+            issue.delete(flush: true)
```
rushtong: Question: Have you tested that this deletes associations (attachments, submissions, extra properties) as well? I've never been very confident of the framework-delegated functions like this.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -217,8 +218,12 @@ class ConsentGroupReview extends Component {
     );
   }
 
-  rejectConsentGroup = (e) => () => {
-
+  rejectConsentGroup() {
+    spinnerService.showAll();
+    ConsentGroup.rejectConsent(this.props.rejectConsentUrl, this.props.consentKey).then(resp => {
+      window.location.href = this.getRedirectUrl(this.props.projectKey);
+      spinnerService.hideAll();
```
rushtong: I didn't catch this before. Could there be an error case where the response fails and `spinnerService.hideAll();` is never invoked? Should this be in a separate handler block? I don't remember the syntax, but there should be an error case for this promise that you could invoke this in.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -327,6 +332,31 @@ class ConsentGroupReview extends Component {
...
+    } else {   
+      let key = projectKey.split("-");
+      let projectType = '';
+      if (key.length === 3) {
+        projectType = key[1].toLowerCase();
+      } else {
+        projectType = key[0].toLowerCase();
+      }
+      return [this.props.serverURL, projectType, "show", projectKey,"?tab=consent-groups"].join("/");
+  }
```
rushtong: Nit: indentation on a few lines in this method is off.

vvicario:
@rushtong I added logical delete for issue associated tables and I addressed your last comments!

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -223,6 +223,9 @@ class ConsentGroupReview extends Component {
     ConsentGroup.rejectConsent(this.props.rejectConsentUrl, this.props.consentKey).then(resp => {
       window.location.href = this.getRedirectUrl(this.props.projectKey);
       spinnerService.hideAll();
+    }).catch(error => {
+      spinnerService.hideAll();
+      console.error(error);
```
rushtong: Nice!

PR #44, opened 2019-01-23
Update issue - Refactor existing code to be used on new flow

vvicario:
## Testing

---

### DIFF
--- a/grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
+++ b/grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
@@ -1,6 +1,5 @@
 package org.broadinstitute.orsp.api
 
-import com.google.gson.Gson
 import grails.converters.JSON
 import grails.rest.Resource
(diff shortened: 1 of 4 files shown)


grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
```diff
@@ -80,15 +79,23 @@ class ProjectController extends AuthenticatedController {
         }
     }
 
-    @Override
-    handleIntake(String key) {
+
+    def update() {
+        Gson gson = new Gson()
+        Map project = gson.fromJson(gson.toJson(request.JSON), Map.class)
```
rushtong: Nit: I've seen this same code in other places. Can we abstract it out to a utility?

rushtong: Also - I'm very suspicious of untyped collections. I strongly prefer that we try to specify types, even if it's `Map<String, Object>` and let `IssueService.updateIssue` handle the particulars of determining what the `Object` part of that is.

PR #56, opened 2019-02-05
Show hide buttons & other Minor Fixes

wloforteBroad:
## Testing

---

### DIFF
--- a/src/main/webapp/projectReview/ProjectReview.js
+++ b/src/main/webapp/projectReview/ProjectReview.js
@@ -44,7 +44,7 @@ class ProjectReview extends Component {
           describeEditType: null,
           editDescription: null,
           projectReviewApproved: 'false'
-        },       
+        },
         fundings: [{ source: { label: '', value: '' }, sponsor: '', identifier: '' }],
         requestor: {
           displayName: '',
@@ -148,7 +148,7 @@ class ProjectReview extends Component {
           prev.futureCopy = futureCopy;
           return prev;
         });
-    });
+      });
   }
 
   getReviewSuggestions() {
@@ -162,10 +162,10 @@ class ProjectReview extends Component {
             return prev;
           });
         } else {
-            this.setState(prev => {
-              prev.reviewSuggestion = false;
-              return prev;
-            });
+          this.setState(prev => {
+            prev.reviewSuggestion = false;
+            return prev;
+          });
         }
       });
   }
@@ -220,13 +220,13 @@ class ProjectReview extends Component {
       })
     );
     if (this.state.reviewSuggestion) {
-        let project = this.getProject();
-        Project.updateProject(this.props.updateProjectUrl, project, this.props.projectKey).then(
-          resp => {
-            this.removeEdits();
-          })
-          .catch(error => {
-            console.error(error);
+      let project = this.getProject();
+      Project.updateProject(this.props.updateProjectUrl, project, this.props.projectKey).then(
+        resp => {
+          this.removeEdits();
+        })
+        .catch(error => {
+          console.error(error);
         });
     }
   }
@@ -234,23 +234,23 @@ class ProjectReview extends Component {
   rejectProject() {
     spinnerService.showAll();
     Project.rejectProject(this.props.rejectProjectUrl, this.props.projectKey).then(resp => {
-      this.setState(prev =>{
+      this.setState(prev => {
         prev.showRejectProjectDialog = !this.state.showRejectProjectDialog;
         return prev;
       });
       window.location.href = [this.props.serverURL, "index"].join("/");
       spinnerService.hideAll();
     })
-    .catch(error => {
-      spinnerService.hideAll();
-      console.error(error);
-    });
+      .catch(error => {
+        spinnerService.hideAll();
+        console.error(error);
+      });
   }
 
   discardEdits() {
     spinnerService.showAll();
     this.removeEdits();
-    this.setState(prev =>{
+    this.setState(prev => {
       prev.showDiscardEditsDialog = !this.state.showDiscardEditsDialog;
       return prev;
     });
@@ -262,7 +262,7 @@ class ProjectReview extends Component {
     Project.updateProject(this.props.updateProjectUrl, project, this.props.projectKey).then(
       resp => {
         this.removeEdits();
-        this.setState(prev =>{
+        this.setState(prev => {
           prev.showApproveDialog = !this.state.showApproveDialog;
           return prev;
         });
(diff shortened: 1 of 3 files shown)


src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -880,38 +916,38 @@ class ProjectReview extends Component {
...
+            disabled: this.state.disableApproveButton,
+            isRendered: this.isAdmin() && projectReviewApproved === 'false' && this.state.readOnly === true
+          }, ["Approve"]),
+
           /*visible for Admin in readOnly mode and if there are changes to review*/
           button({
             className: "btn buttonPrimary floatRight",
             onClick: this.handleApproveDialog,
-            isRendered: this.isAdmin() && this.state.reviewSuggestion
+            isRendered: this.isAdmin() && this.state.reviewSuggestion && (this.state.readOnly === true && projectReviewApproved === 'true')
```
dags: not needed parentheses in *(this.state.readOnly === true && projectReviewApproved === 'true')* on isRendered ... can be removed.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -209,37 +209,48 @@ class ProjectReview extends Component {
     return value === '' || value === null || value === undefined;
   }
 
-  approveRevision = (e) => () => {
+  approveRevision = () => {
     this.setState({ disableApproveButton: true })
-    const data = { projectReviewApproved: true }
+    const data = { projectReviewApproved: 'true' }
```
rushtong: In general, I prefer booleans to really be booleans. I'm OK with this change for this PR, but in the future, we should try to keep the types as clean as possible.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -880,38 +916,38 @@ class ProjectReview extends Component {
...
+            disabled: this.compareObj("formData", "current") || this.compareObj("formData", "editedForm"),
             isRendered: this.state.readOnly === false
           }, ["Submit Edits"]),
 
+          /*visible for Admin in readOnly mode and if the project is in "pending" status*/
+          button({
+            className: "btn buttonPrimary floatRight",
+            onClick: this.handleApproveInfoDialog,
+            disabled: this.state.disableApproveButton,
+            isRendered: this.isAdmin() && projectReviewApproved === 'false' && this.state.readOnly === true
```
rushtong: This is an example of why I prefer a typed boolean approach. This could (and probably should) be simplified to 
```suggestion
            isRendered: this.isAdmin() && !projectReviewApproved && this.state.readOnly
```

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -458,9 +469,9 @@ class ProjectReview extends Component {
       prev.formData.projectExtraProps[field] = value;
       return prev;
     },
-    () => {
-      if(this.state.errorSubmit == true) this.isValid() 
-    }); 
+      () => {
+        if (this.state.errorSubmit == true) this.isValid()
```
rushtong: Can't this be simplified ? (here and elsewhere)
```suggestion
        if (this.state.errorSubmit) this.isValid()
```

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -514,39 +539,40 @@ class ProjectReview extends Component {
...
-   let subjectProtectionError = false;
-   let editTypeError = false;
-   let editDescriptionError = false;
-   
-   if (this.state.projectType === "IRB Project" && this.isEmpty(this.state.formData.projectExtraProps.editDescription)) {
+    let descriptionError = false;
+    let projectTitleError = false;
+    let subjectProtectionError = false;
+    let editTypeError = false;
+    let editDescriptionError = false;
```
rushtong: There are a *lot* of white space changes. It would be simpler to review if we split those out into a single PR. I'm fine with this for this PR, but for the future, it would be easier to review :-)

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -555,23 +581,31 @@ class ProjectReview extends Component {
           show: this.state.showRejectProjectDialog,
           handleOkAction: this.rejectProject,
           title: 'Remove Project Confirmation',
-          bodyText: 'Are you sure yo want to remove this project?',
+          bodyText: 'Are you sure you want to remove this project?',
```
rushtong: Nice catch ... I should have seen these.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -732,28 +767,29 @@ class ProjectReview extends Component {
...
+        Panel({ title: "Notes to ORSP", isRendered: this.state.readOnly === false }, [
+          div({ isRendered: this.projectType === "IRB Project" }, [
+            InputFieldRadio({
+              id: "radioDescribeEdits",
+              name: "describeEditType",
+              currentValue: this.state.current.projectExtraProps.describeEditType,
+              currentOptionLabel: this.state.current.projectExtraProps.describeEditType === 'newAmendment' ?
+                "I am informing Broad's ORSP of a new amendment I already submitted to my IRB of record" :
+                "I am requesting assistance in updating and existing project",
+              label: "Please choose one of the following to describe the proposed Edits: ",
```
rushtong: nit: should `Edits` be `edits`?

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -880,38 +916,38 @@ class ProjectReview extends Component {
...
           /*visible for Admin in readOnly mode and if the project is in "pending" status*/
           button({
-            className: "btn buttonPrimary floatRight",
-            onClick: this.approveRevision(),
-            disabled: this.state.disableApproveButton,
-            isRendered: this.isAdmin() && !this.state.formData.projectExtraProps.projectReviewApproved
-          }, ["Approve"]),
+            className: "btn buttonSecondary floatRight",
+            onClick: this.handleRejectProjectDialog,
+            isRendered: this.isAdmin() && projectReviewApproved === 'false' && this.state.readOnly === true
```
rushtong: Similar feedback to above, this would read a lot better as 
```suggestion
            isRendered: this.isAdmin() && !projectReviewApproved && this.state.readOnly
```

src/main/webapp/projectReview/ProjectReview.js
wloforteBroad: Sorry for that, Nadya did some code format in her commit. Will try not to do this again :)

src/main/webapp/projectReview/ProjectReview.js
dags: We agree on that. This value comes from key / value pair in issue_extra_properties where both key and value are stored as string, since could hold any data type serialized as string. We retrieve all  extra properties at once for an issue.

src/main/webapp/projectReview/ProjectReview.js
wloforteBroad: yes it should, I'll change it

src/main/webapp/projectReview/ProjectReview.js
rushtong: No worries, thank you!

PR #59, opened 2019-02-08
BTRX - Project Review - Fundings

LeoForconesi:
## Changes
- Implement funding review, edition and compare in project review.
- Fix elements comparison order in Multi select component.
- Correct console warning for uncontrolled variable.

## Notes
- You will find a couple of `todo` flags, these indicate some improvements coming in the next PR for saving code and complexity.

---


src/main/webapp/components/Fundings.js
```diff
@@ -111,11 +184,14 @@ export const Fundings = hh(class Fundings extends Component {
                       index: idx,
                       name: "source",
                       options: fundingOptions,
-                      value: this.props.fundings[idx].source,
+                      value: this.props.edit ? rd.future.source : rd.source,
+                      currentValue: this.props.edit ? current[idx].current.source : rd.source,
                       onChange: this.handleFundingSelect,
-                      error: this.props.error && idx === 0,
+                      error: this.props.edit === true  && this.props.errorIndex !== null? this.props.error && this.props.errorIndex.includes(idx) : this.props.error && idx === 0,
```
rushtong: To be 100% honest, these kinds of multi-conditional checks make my eyes cross. This seems like a good place to abstract the logic into a function that is easier to grok.

src/main/webapp/project/NewProjectGeneralData.js
```diff
@@ -216,7 +216,7 @@ export const NewProjectGeneralData = hh(class NewProjectGeneralData extends Comp
             id: "inputStudyActivitiesDescription",
             name: "studyDescription",
             label: "Describe Broad study activities* ",
-            moreInfo: "(briefly, in 1-2 paragraphs, with attention to wheter or not protected health information will be accessed, future data sharing plans, and commercial or academic sample/data sources. For commercially purchased products, please cite product URL.)",
+            moreInfo: "(briefly, in 1-2 paragraphs, with attention to whether or not protected health information will be accessed, future data sharing plans, and commercial or academic sample/data sources. For commercially purchased products, please cite product URL.)",
```
rushtong: Nice catch

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -45,14 +46,19 @@ class ProjectReview extends Component {
...
+        }],
         requestor: {
           displayName: '',
           emailAddress: ''
         }
       },
       disableApproveButton: false,
       reviewSuggestion: false,
+      futureCopy: {
+      },
```
rushtong: nit: can be a single line: `futureCopy: {},`

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -330,19 +354,38 @@ class ProjectReview extends Component {
...
       });
     }
     return fundingList;
   }
 
   compareObj(obj1, obj2) {
-    return JSON.stringify(this.state[obj1]) === JSON.stringify(this.state[obj2]);
+    let form1 = JSON.parse(JSON.stringify(this.state[obj1]));
+    let form2 = JSON.parse(JSON.stringify(this.state[obj2]));
+    if ( form1 !== undefined && form2 !== undefined && form1.fundings !== undefined && form2.fundings !== undefined) {
```
rushtong: Could you do something like this instead? (https://www.w3schools.com/jsref/jsref_every.asp)
```
[form1, form2, form1.fundings, form2.fundings].every(function(el){ return el !== undefined })
```

rushtong: Alternatively, we could use lodash.get which has a form of null-safe traversal: https://lodash.com/docs/#get

rushtong: I admit to being spoiled by Groovy's null-safe handling - it's one of the things that keeps me coming back to languages like it and Kotlin.

src/main/webapp/components/Fundings.js
LeoForconesi: This is now corrected, I must have overlooked it...

src/main/webapp/projectReview/ProjectReview.js
LeoForconesi: I really liked lodash approach, so I've imported it to the project. Thanks for the suggestion! now we can count on this library.

PR #66, opened 2019-02-19
Btrx 554 Implement Data Use Letter electronic Form (UI)

vvicario:
## Testing

---


grails-app/views/dataUseLetter/index.gsp
```diff
@@ -0,0 +1,26 @@
...
+    };
+    </script>
+</head>
+<body>
+
+<div id="dataUseLetter"></div>
+<asset:javascript src="dataUseLetter.js"/>
+
+</body>
+</html>
```
rushtong: Not for this PR, but we should probably create a new template layout specifically for these react pages. Something like `<meta name="layout" content="main"/>` so we can start consolidating common static urls and script references.

src/main/webapp/components/InputField.js
```diff
@@ -15,10 +15,10 @@ export const InputField = hh(class InputField extends Component {
   }
 
   render() {
-    const { value, label, error, errorMessage, moreInfo, children, readOnly, currentValue = null, currentValueStr, edited = false } = this.props;
+    const { value, label, additionalClass, error, errorMessage, moreInfo, children, readOnly, currentValue = null, currentValueStr, edited = false } = this.props;
```
rushtong: There's definitely a better way to handle this in react, but I need to explore it more to get a better handle on how to explain it. In essence, style properties should be mergeable down from parent components so that you don't necessarily have to pull them explicitly out of props. I'll see if I can dig up an example.

nlopezzalba: @rushtong  what I intended to do here was to provide the possibility to add a class (any) from the component that is calling this component (in this case the class "inputFieldCkb" from 
InputFieldCheckbox to InputField). Since InputField handles several components, I only need it to apply certain class if InputFieldCheckbox is calling it. Don't see how the solution you mention above can help me solve that issue, but I'm open for suggestions of course.

rushtong: Thanks Nadya, I agree - the solution I mentioned above doesn't solve this case - I'm fine with what you have here. I was hoping to be better able to explain it, but cannot at the moment :-(  If I do come up with a better way to phrase what I'm thinking, I'll definitely do that.

PR #68, opened 2019-02-22
[BTRX-558] Send link to DUL form

tadeork:
## Addresses
Jira story https://broadinstitute.atlassian.net/browse/BTRX-558
Send email to a collaborator when Data Use Letter is selected. 
Show spinner when uploading file.
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/DulNotifyController.groovy
```diff
@@ -0,0 +1,32 @@
...
+import org.broadinstitute.orsp.User
+
+@Slf4j
+class DulNotifyController extends AuthenticatedController{
+
+    def sendNotifications() {
+        User user = userService.findUser(request.JSON['userName'].toString())
+        def issue = queryService.findByKey(params.consentKey)
+        def usersNotif = [request.JSON['recipients']]
+//        def dulLinksForm = method.getLinks()
```
rushtong: nit: rm

grails-app/views/notify/dulFormLink.gsp
```diff
@@ -0,0 +1,9 @@
+<p>
+Hello,
+This email is a request from your collaborator at the Broad Institute to have your IRB or Ethics Committee of record
+complete the linked <b><i>Data Use Letter</i></b> form by interpreting the consent form for this study.
+Could you please share <a href="${details}">this link</a> and/or email with a member of your IRB or Ethics Committee so they can proceed with filling out and submitting the Data Use Letter form?
+If you have any questions, please do not hesitate to email your collaborator (cc'd) or ORSP@broadinstitute.org.
+Thank you,
+Sent on behalf of Broad Institute's Office of Research Subject Protection (ORSP)
+</p>
```
rushtong: nit: without breaks, this won't format in the way it looks here. Just want to make sure it really looks the way you want it to. There is a mail group that all email goes through that you can get access to (will post in slack) so you can see what it looks like for real.

src/main/webapp/components/AddDocumentDialog.js
```diff
@@ -94,6 +105,43 @@ export const AddDocumentDialog = hh(class AddDocumentDialog extends Component {
     });
   };
 
+  validEmail = (email) => {
```
rushtong: This should be a generalized function so we can keep the regex in one spot. I can see this being copy/pasted unnecessarily.

PR #69, opened 2019-02-25
BTRX-565 - Integrate document modal and DUL form. Implement Submit DUL form

vvicario:
## Testing

---


grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
```diff
@@ -0,0 +1,48 @@
+package org.broadinstitute.orsp
+
+import grails.gorm.transactions.Transactional
+
+class DataUseLetterService {
+    QueryService queryService
+
+    String generateDul(Object input) {
```
rushtong: nit: It would prefer if we could have a better type for this. That would allow you to avoid things like `input["user"] as String`. Groovy lets you do this kind of dynamic typing, but it generally leads to bugs in the long run.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseLetterController.groovy
```diff
@@ -0,0 +1,56 @@
...
+import org.broadinstitute.orsp.utils.IssueUtils
+
+@Slf4j
+@Resource(readOnly = false, formats = ['JSON', 'APPLICATION-MULTIPART'])
+class DataUseLetterController extends AuthenticatedController {
+
+
+    @Override
+    def create () {
+        Object input = IssueUtils.getJson(Map.class, request.JSON)
```
rushtong: Here is where I would try to deconstruct the object into a more useful thing.

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+        if (newDul.hasErrors()) {
+            throw new DomainException(newDul.getErrors())
+        } else {
+            newDul.uid
+        }
+    }
+
+    @Transactional
+    def udpateDataUseLetter(Object input) {
```
rushtong: Same feedback as above. You can avoid casting `input.dulInfo as String` if there was a proper object here to refer to.

LeoForconesi:
@rushtong 
Corrected all generic `Object` cast into `DataUseLetter` model.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseLetterController.groovy
```diff
@@ -0,0 +1,58 @@
...
+            render([dulToken: newDul.getUid()] as JSON)
+        } catch(Exception e) {
+            response.status = 500
+            render([error: e.message] as JSON)
+        }
+    }
+
+    @Override
+    def update () {
+        DataUseLetter input = IssueUtils.getJson(DataUseLetter.class, request.JSON)
```
rushtong: Thank you!

PR #74, opened 2019-02-27
Project Review Improvements

wloforteBroad:
https://broadinstitute.atlassian.net/browse/BTRX-570
---

### DIFF
--- a/src/main/webapp/components/InputFieldText.js
+++ b/src/main/webapp/components/InputFieldText.js
@@ -1,6 +1,8 @@
 import { Component } from 'react';
 import { input, hh, div } from 'react-hyperscript-helpers';
 import { InputField } from './InputField';
+import { areSomeTheseThingsTruthy } from '../util/Utils';
+import { isEmpty } from '../util/Utils';
 import './InputField.css';
 
 export const InputFieldText = hh(class InputFieldText extends Component {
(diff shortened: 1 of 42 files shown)


src/main/webapp/components/InputFieldText.js
```diff
@@ -8,8 +8,9 @@ export const InputFieldText = hh(class InputFieldText extends Component {
   render() {
 
     const { value, currentValue } = this.props;
-    const edited = value !== currentValue && currentValue !== undefined || this.props.valueEdited === true;
-
+    const edited = value !== currentValue && currentValue !== undefined
+      || this.props.valueEdited === true
+      || this.props.edit === true && value !== currentValue && currentValue === undefined && value !== '';
```
rushtong: Ugh, this conditional is horrible. I don't have a better suggestion than this, so I'm fine with this. It might be nice to be able to abstract these kinds of checks into a utility function. Something that checks for truthiness on a collection of objects, like `const edited = areAllTheseThingstruthy() { arguments.forEach(checkTruthiness(thing));}`

rushtong: Maybe something like this? https://lodash.com/docs/4.17.11#every

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -126,28 +128,10 @@ class ProjectReview extends Component {
...
-          formData = JSON.parse(currentStr);
-          current = JSON.parse(currentStr);
-          future = JSON.parse((currentStr));
-          futureCopy = JSON.parse(currentStr);
-        }
-
+        formData = JSON.parse(currentStr);
+        current = JSON.parse(currentStr);
+        future = JSON.parse((currentStr));
+        futureCopy = JSON.parse(currentStr);
```
rushtong: Nice simplification

src/main/webapp/components/InputFieldText.js
vvicario: Agree! I created some generic validations that will be util

PR #82, opened 2019-03-06
BTRX - international Cohorts fix

LeoForconesi:
## Changes
- Before we build our object to be sent and save after approval, we needed a way to refresh our variable `intCohortsAnswers`, which if the user did never edited these values,  will remain empty. 
For this we created `parseIntCohorts( )` that will take all answers scattered on extra properties coming from 'Issue_review' database table, and resfresh our oject.
---


grails-app/assets/javascripts/consentGroupReview.js
```diff
@@ -91491,6 +91490,22 @@ var ConsentGroupReview = function (_Component) {
       });
       this.init();
     }
+  }, {
+    key: 'parseIntCohorts',
+    value: function parseIntCohorts() {
+      var _this3 = this;
```
rushtong: the `var`s should be `let`s

grails-app/assets/javascripts/consentGroupReview.js
```diff
@@ -91513,21 +91528,21 @@ var ConsentGroupReview = function (_Component) {
   }, {
     key: 'isCurrentUserAdmin',
     value: function isCurrentUserAdmin() {
-      var _this3 = this;
+      var _this4 = this;
```
rushtong: I'm curious about why we need to rename these. Are we using these in local storage?

LeoForconesi:
@rushtong I didn't change any of those files intentionally, it seems that the reason of those differences is the re-compiled files produced by  `webpackDev`. I'm not sure if I can fix them, will try though.

rushtong:
@LeoForconesi - Ah ... I missed that detail. If you run the webpackProd command, my comments should be ignored. Looks like webpack made the `var`s too.

PR #83, opened 2019-03-07
BTRX - 551 - Consent group associated projects

LeoForconesi:
## [Addresses
BTRX-551](BTRX-551-consentGroupAssociatedProjects) - Consent group associated Project

## Changes
- Implement unlink and show consent group, associated projects

---


grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -127,4 +127,36 @@ class NewConsentGroupController extends AuthenticatedController {
...
+    /**
+     * This action breaks the link between a project and the consent from the point of view of the consent.
+     *
+     */
+    def unlinkAssociatedProjects () {
+        Issue issue = queryService.findByKey(params.consentKey)
+        Issue target = queryService.findByKey(request.JSON["projectKey"] as String)
+        def links = ConsentCollectionLink.findAllByProjectKeyAndConsentKey(target.projectKey, issue.projectKey)
+        deleteCollectionLinks(links)
+        response.status = 200
```
rushtong: Note that this will always be a 200 even if the delete fails. It might make more sense return a boolean or exception from the delete call and then use that to craft the response status.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -127,4 +127,36 @@ class NewConsentGroupController extends AuthenticatedController {
...
+        render([collectionLinks: collectionLinks.linkedProject] as JSON)
+        collectionLinks
+    }
+
+    /**
+     * This action breaks the link between a project and the consent from the point of view of the consent.
+     *
+     */
+    def unlinkAssociatedProjects () {
+        Issue issue = queryService.findByKey(params.consentKey)
```
rushtong: really small naming nit here and in `getConsentCollectionLinks`. Since this is finding a consent group, it should be named as such
```suggestion
        Issue consent = queryService.findByKey(params.consentKey)
```

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -127,4 +127,36 @@ class NewConsentGroupController extends AuthenticatedController {
         response.status = 200
         render([consent: consent] as JSON)
     }
+
+    Collection<ConsentCollectionLink> getConsentCollectionLinks() {
+        Issue issue = queryService.findByKey(params.consentKey)
+        Collection<ConsentCollectionLink> collectionLinks = queryService.findCollectionLinksByConsentKey(issue.projectKey)
+        render([collectionLinks: collectionLinks.linkedProject] as JSON)
```
rushtong: Should we need to handle a 404 here? It would be nice for completeness.

src/main/webapp/components/Table.js
```diff
@@ -52,6 +52,23 @@ export const Table = hh(class Table extends Component {
...
+  unlinkProject = (row) => {
+    return button({
+      className: "btn btn-xs",
+      onClick : this.props.unlinkProject(row),
+      disabled: !this.props.isAdmin
+    }, ["Unlink"])
+  };
+
+  redirectToProject = (cell, row) => {
+    let url = this.props.handleRedirectToProject(row.projectKey);
```
rushtong: can this be a const?

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
LeoForconesi: I'm not sure if we want to handle a 404, this method is expected to return something or an empty collection. 
Since it's only used for listing Links this would need us modify the Ui to ignore this 404. If this doesn't rise any concerns to you,  i'll gladly add it.

rushtong: Makes sense then 👍

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
LeoForconesi: You're right, if any error occurs now it will return a 500

LeoForconesi:
@rushtong If you have some time please, check these changes.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -63,6 +65,35 @@ class ConsentGroupDocuments extends Component {
...
+    });
+  };
+
+  redirectToProject = (projectKey) => {
+    let key = projectKey.split("-");
+    let projectType = '';
+    if (key.length === 3) {
+      projectType = key[1].toLowerCase();
+    } else {
+      projectType = key[0].toLowerCase();
```
rushtong: I do think that splitting the project key string is a brittle way to get the project type. In groovy, we have code that figures out what the controller is, but I don't think we have anything in javascript for this. It might be a little overkill, but if we had an API method that said, "get me the projectType for this projectKey", this would be a much more reliable method. 

Not a requirement for this PR though.

LeoForconesi:
I've created a technical debt task for "create an API for getting the project type"  so we can keep track. 
[BTRX-591](https://broadinstitute.atlassian.net/browse/BTRX-591)

PR #86, opened 2019-03-08
[BTRX-550 dur

tadeork:
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -128,6 +132,12 @@ class NewConsentGroupController extends AuthenticatedController {
         render([consent: consent] as JSON)
     }
 
+    def findUseRestriction() {
```
rushtong: nit: I think `getDataUseRestriction` is more idiomatic of our naming style.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -128,6 +132,12 @@ class NewConsentGroupController extends AuthenticatedController {
...
+        def restriction = DataUseRestriction.findByConsentGroupKey(params.consentKey)
+        Collection<String> duSummary = consentService.getSummary(restriction)
```
rushtong: This can fail, i.e., return nothing if the consent doesn't have one created for it yet.

src/main/webapp/components/Documents.js
```diff
@@ -54,7 +66,21 @@ export const Documents = hh(class Documents extends Component {
     this.setState({ showAddAdditionalDocuments: !this.state.showAddAdditionalDocuments });
   };
 
+  findDul = () => {
+    let dulPresent = false;
+    if (this.props.keyDocuments.length !== 0) {
+      this.props.keyDocuments.forEach(docs => {
+        if (docs.fileType === 'Data Use Letter') {
```
rushtong: It would be nice if `'Data Use Letter'` were an enum so we could abstract this kind of check. Could easily be a tech-debt story if you think it's a good idea.
https://stackoverflow.com/questions/287903/what-is-the-preferred-syntax-for-defining-enums-in-javascript

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -65,6 +67,20 @@ class ConsentGroupDocuments extends Component {
...
+    ConsentGroup.getUseRestriction(this.props.useRestrictionUrl, this.props.projectKey).then(resp => {
+      this.setState(prev => {
+        prev.restriction = resp.data.restriction;
+        if (resp.data.restrictionId !== null) {
+          prev.restrictionId = resp.data.restrictionId.id;
+        } else {
+          prev.restrictionId = null;
+        }
+        return prev;
+      })
```
rushtong: I think it would be a lot safer to take this if out of the setState call. You have `resp.data.restrictionId` in `resp`, so something like this might work:
```
const newRestrictionId = resp.data.restrictionId ? resp.data.restrictionId.id : null;
this.setState(prev => { 
    prev.restrictionId = newRestrictionId; 
    return prev;
});
```

PR #87, opened 2019-03-11
BTRX - 577 - Add International Cohort Tab to project creation flow

LeoForconesi:
## Addresses
[BTRX-577](https://broadinstitute.atlassian.net/browse/BTRX-577) : ORSP: Add International Cohort Tab to NE and NHSR Projects creation flow

## Changes
- International Cohorts component is moved to be re-used by NewProject and NewConsentGroup.
- Some minor code refactoring
- We've included transpiled javascripts from webpackProd
---


src/main/webapp/project/NewProject.js
```diff
@@ -143,6 +150,16 @@ class NewProject extends Component {
           }
         });
     }
+
+    let internationalCohortsQuestions = this.state.intCohortsDetermination.questions;
+    if (internationalCohortsQuestions !== null && internationalCohortsQuestions.length > 1) {
+      internationalCohortsQuestions.map((q, idx) => {
+        if (q.answer !== null) {
```
rushtong: Do you care about undefined here?

LeoForconesi: Actually no, since this variable is defined in the component's state, from its constructor.
As a matter of fact, validate to null would be also useless, this array can only be an empty array, or the set of questions when InternationalCohorts component is initialized when a change occurs, and updates the parent state by it. If its not completely clear, please ping me on slack

PR #89, opened 2019-03-12
BTRX - 578 - Add info security tab in Project creation

LeoForconesi:
## Addresses
[BTRX-578](https://broadinstitute.atlassian.net/browse/BTRX-578) : ORSP: Add InfoSecurity Tab to NE and NHSR Projects creation flow

## Changes
- Move DataSharing component that originally was used only by consent group, to a common place to be re-used by Project.
- Move validation into DataSharing component
- Rename multiple variables to more representative in both Consent group and Project components
- Code Refactoring.

---


src/main/webapp/components/Security.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+    if (isEmpty(this.state.formData.accessible)) {
+      accessible = true;
+      isValid = false;
+    }
+    if (!isEmpty(this.state.formData.accessible) && this.state.formData.accessible === "true" && isEmpty(this.state.formData.textAccessible)) {
+      textAccessible = true;
+      isValid = false;
+    }
+    if (field === undefined || field === null || field === 3) {
```
rushtong: Can you explain where `field === 3` comes from? I think a comment here that explains why you need this check would be helpful.

src/main/webapp/components/Security.js
```diff
@@ -0,0 +1,278 @@
...
+        prev.errors.sensitive = sensitive;
+        prev.errors.accessible = accessible;
+        prev.errors.textCompliance = textCompliance;
+        prev.errors.textSensitive = textSensitive;
+        prev.errors.textAccessible = textAccessible;
+        return prev;
+      });
+    }
+    else if (field === 'pii' || field === 'compliance' || field === 'textCompliance' || field === 'sensitive'
+      || field === 'textSensitive' || field === 'accessible' || field === 'textAccessible') {
```
rushtong: These field comparisons (here and below) seem very brittle to me. It looks like there's a 1 to 1 alignment with both the `state.formData` and `state.errors` properties. The comparisons are also very repetitive. It would be nice if we could, instead, say something like this (warning, pseudo-code):

```
else if (fieldNames.contains(field)) {
  this.setState(prev => {
    fieldNames.each { f -> prev.errors.get(f) = f }
    return prev
});
```
That would save a whole lot of boilerplate code here and below. I don't have a great idea about how to do this in javascript, but maybe if `state.formData` and `state.errors` were generalized into an object, that might make it easier.

src/main/webapp/project/NewProject.js
```diff
@@ -496,6 +528,17 @@ class NewProject extends Component {
           errors: this.state.showErrorIntCohorts,
           origin: 'newProject'
         }),
+        Security({
+          title: "Security",
+          step: 3,
```
rushtong: Ah ... this looks like the answer to my previous question about checking against `3`. The connection between here and there is non-obvious and definitely needs some comments about it.

LeoForconesi: As matter a fact, this has nothing to do with the number 3 you saw above. I checked the entire code trying to understand its usage (if any) and couldn't find one... Since this code had some simplification on its usage but at the same time I tried to keep at minimum too much changes, it seems that this is old code now unused. 
I'll keep an eye to clean up this type of orphan code in the rest of the project as I continue refactoring stuff.

src/main/webapp/components/Security.js
LeoForconesi: Removed `field === 3` since it's unused

src/main/webapp/components/Security.js
LeoForconesi: I hope I have not been too drastic on this. 
I the midle of doing this change I saw it was not neccesary anymore to check by each specific field. This was used on the update functions above, but I think it is unnecessary now. 

If you see the entire validate function, you will see it first validates all fields in the form, to later after go to this portion of  the code, where it would only update that specific error flag in the state... functionality will not change  if we just update all flags in the state at once, without caring individual fields, and we save logic and code in the process.
Does it make sense for you?

rushtong: It does make sense, thank you!

src/main/webapp/project/NewProject.js
rushtong: Nice ... really glad I actually found something :-)

PR #90, opened 2019-03-12
[BTRX-580] and [BTRX-585] Clarification Request

tadeork:
## Addresses
Dialog to request clarification in the Review tab of Project and Consent Group. This will send emails to the requestor, PMs depending on the case.
https://broadinstitute.atlassian.net/browse/BTRX-580 and https://broadinstitute.atlassian.net/browse/BTRX-585
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -0,0 +1,55 @@
...
+import grails.rest.Resource
+import org.broadinstitute.orsp.AuthenticatedController
+import org.broadinstitute.orsp.Comment
+import org.broadinstitute.orsp.Issue
+import org.broadinstitute.orsp.IssueType
+import org.broadinstitute.orsp.NotifyArguments
+
+
+@Resource(readOnly = false, formats = ['JSON', 'APPLICATION-MULTIPART'])
+class ClarificationController extends AuthenticatedController{
```
rushtong: nit: spacing
```suggestion
class ClarificationController extends AuthenticatedController {
```

grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -0,0 +1,55 @@
...
+            } else {
+                toAddresses.addAll(userService.findUser(issue.getReporter())?.collect {it.emailAddress})
+            }
+            // Extra check to cover error condition of empty PMs
+            if (toAddresses.isEmpty()) {
+                log.error("Issue " + issue.projectKey + " has empty PMs.")
+                toAddresses.addAll(notifyService.getOrspSpecialRecipients())
+            }
+
+            notifyService.sendClarificationRequest(
```
rushtong: The return value of this comes from `NotifyService. sendMail`  and might contain an error that gets lost here.

grails-app/views/notify/clarificationRequest.gsp
```diff
@@ -0,0 +1,11 @@
...
+    <p>
+        User ${user.displayName}, has requested a clarification  for ${issue.typeLabel} ${issue.projectKey}
+    </p>
+    <p>Clarification :</p>
+    <p style="margin-left: 50px;">${raw(comment)}</p>
+    <p>
+        Details regarding this submission can be accessed at:
+        <a href="${issueLink}">${issueLink}</a>
+    </p>
+</div>
```
rushtong: Just a note that I'm glad this pattern met the need for this feature. In retrospect, we could have done something similar to what Consent does with Freemarker templates. In the long run, we probably do want to get away from using GSP files completely, but for now, this is great.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -588,6 +592,17 @@ class ConsentGroupReview extends Component {
...
 
+  requestClarification = () => {
+    this.setState({
+      showRequestClarification: !this.state.showRequestClarification
+    });
+  };
+
+  closeClarificationModal = () => {
+    this.setState({ showRequestClarification: !this.state.showRequestClarification });
+  };
```
rushtong: Here and above - these need to use the updater form because the toggle depends on current state. See the [react docs for an example](https://reactjs.org/docs/react-component.html#setstate). 

>  If the next state depends on the current state, we recommend using the updater function form, instead:
```
this.setState((state) => {
  return {quantity: state.quantity + 1};
});
```

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -519,18 +525,18 @@ class ProjectReview extends Component {
...
-      this.setState({
-        showApproveDialog: !this.state.showApproveDialog,
-        editedForm: {},
-        errorSubmit: false
-      });
+    this.setState({
+      showApproveDialog: !this.state.showApproveDialog,
+      editedForm: {},
+      errorSubmit: false
+    });
```
rushtong: Here is another case where a current state calculation depends on previous state so it should use the updater form.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -545,6 +551,16 @@ class ProjectReview extends Component {
...
 
+  requestClarification = () => {
+    this.setState({
+      showRequestClarification: !this.state.showRequestClarification
+    });
+  };
+
+  closeClarificationModal = () => {
+    this.setState({ showRequestClarification: !this.state.showRequestClarification });
+  };
```
rushtong: These state updaters are duplicated here and in `ConsentGroupReview.js`. It would be nice if we could abstract these into one place.

PR #95, opened 2019-03-14
Btrx 413 466 permissions

dags:
## Addresses

**BTRX-413: In Search Results, display all results post-permissioning but only link results which user can access**

**BTRX-466: Permissions: Add user type which can VIEW=ALL RECORDS but EDIT=NONE**

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/IssueListController.groovy
```diff
@@ -19,10 +19,15 @@ class IssueListController extends AuthenticatedController {
         if (isORSP()) {
             users.add(SupplementalRole.ORSP)
         }
+
         if (isComplianceOffice()) {
             users.add(SupplementalRole.COMPLIANCE_OFFICE)
         }
 
+        if (isReadOnlyAdmin()) {
+            users.add(SupplementalRole.RONLY_ADMIN)
```
rushtong: Nit: naming, I would prefer that this be spelled out, `RONLY` looks a bit like `RONY` which is short for pepperoni pizza in these parts 😄 
```suggestion
            users.add(SupplementalRole.READ_ONLY_ADMIN)
```

grails-app/controllers/org/broadinstitute/orsp/SearchController.groovy
```diff
@@ -7,11 +7,12 @@ import org.grails.plugins.web.taglib.ApplicationTagLib
 import java.text.SimpleDateFormat
 
 @SuppressWarnings("GroovyAssignabilityCheck")
-class SearchController {
+class SearchController implements UserInfo {
```
rushtong: Really glad this pattern worked for you - I'd like to make more use of traits in this codebase.

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -142,6 +142,8 @@ class Issue implements LogicalDelete<Issue> {
 
     transient Collection<String> getPMs() { getExtraProperties().findAll { it.name == IssueExtraProperty.PM }.collect { it.value } }
 
+    transient Collection<String> getCollaborators() { getExtraProperties().findAll { it.name == "collaborator" }.collect { it.value } }
```
rushtong: Nit: We have a string property for this: `IssueExtraProperty.COLLABORATOR`
Also, we have some custom collaborator logic in `Issue.getExtraPropertiesMap()` that I'm having trouble following. @LeoForconesi - Can you elaborate here on if we're missing anything? I think we could use some comments around why `getExtraPropertiesMap()` is doing what it is doing if we have a `Issue.getCollaborators()` method.

grails-app/domain/org/broadinstitute/orsp/SupplementalRole.groovy
```diff
@@ -5,6 +5,7 @@ class SupplementalRole {
     public static final String COMPLIANCE_OFFICE = "Compliance Office"
     public static final String ORSP = "orsp"
     public static final String ADMIN = "admin"
+    public static final String RONLY_ADMIN = "ro_admin"
```
rushtong: nit: See my previous RONY comment.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
+package org.broadinstitute.orsp
+
+import grails.gorm.transactions.Transactional
+import groovy.util.logging.Slf4j
+
+@Transactional
+@Slf4j
+class PermissionService implements UserInfo{
```
rushtong: nit spacing
```suggestion
class PermissionService implements UserInfo {
```

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
...
+
+    UserService userService
+
+    // get issue's collaborators as a List<String>
+    def getIssueCollaborators(issue) {
```
rushtong: I would prefer that these `def`s instead be the return type. Also, the argument should also be typed:
```suggestion
    List<String> getIssueCollaborators(Issue issue) {
```
Here and throughout.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
...
+        def collaborators = issue.extraProperties.findAll ({ it.name == 'collaborator' }).collect { property -> property.value }
+        collaborators
```
rushtong: This can be a one-liner for simplicity: 
```suggestion
List<String> getIssueCollaborators(issue) {
        issue.extraProperties.findAll ({ it.name == 'collaborator' }).collect { property -> property.value }
}
```
Here and throughout.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // get issue's pis as a List<String>
+    def getIssuePIs(issue) {
+        def pis = issue.extraProperties.findAll ({ it.name == 'pi' }).collect { property -> property.value }
+        pis
+    }
+
+    // get issue's actors as a List<String>
+    def getIssueActors(issue) {
```
rushtong: This looks unused.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
rushtong: Just a note on typing. Without `issue` being typed, the IDE can't determine the type of `issue.extraProperties` so it ends up being dynamically typed. Note in this screenshot how the IDE doesn't know what `property.value` is 
Screen Shot 2019-03-15 at 9 25 32 AM

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
...
+        def actors = issue.extraProperties.findAll ({ it.name == 'actor' }).collect { property -> property.value }
+        actors
+    }
+
+    // verifies if logged user belogns to some user list ....
```
rushtong: nit: typo
```suggestion
    // verifies if logged user belongs to some user list ....
```

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,50 @@
...
+        if (issue.reporter == userName
+                || getIssueCollaborators(issue).indexOf(userName) >= 0
+                || getIssuePMs(issue).indexOf(userName) >= 0
+                || getIssuePIs(issue).indexOf(userName) >= 0
+                || isAdmin
+                || isReadOnlyAdmin) {
+            return false
+        }
+        true
+    }
```
rushtong: This took me some time to grok as it's doing inverse logic. To make this easier to spot more quickly, can I suggest that we define a named variable to the result of the if and then return the negation of that? Something along the lines of:
```
        boolean userHasAccess = (issue.reporter == userName
                || getIssueCollaborators(issue).indexOf(userName) >= 0
                || getIssuePMs(issue).indexOf(userName) >= 0
                || getIssuePIs(issue).indexOf(userName) >= 0
                || isAdmin
                || isReadOnlyAdmin)
        !userHasAccess
```

rushtong: Alternatively, instead of checking for `forbidden-ness`, can we instead check for `allowed access` ? I think that would make the code a bit more readable.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -651,11 +651,12 @@ class QueryService implements Status {
...
         def ids = rows.collect{it.get("id")}
+        Collection result = Collections.emptyList()
+
         if (ids.size() > 0) {
-            Issue.findAllByIdInList(ids)
-        } else {
-            Collections.emptyList()
+            result = Issue.findAllByIdInList(ids)
         }
+        return result;
```
rushtong: Nit: no need for return here:
```suggestion
        result;
```

orsp-pub.iml
```diff
@@ -1,7 +1,13 @@
...
+<module external.linked.project.id="orsp-pub" external.linked.project.path="$MODULE_DIR$" external.root.project.path="$MODULE_DIR$" external.system.id="GRADLE" external.system.module.group="org.broadinstitute" external.system.module.version="3.9.7" type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/.gradle" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/out" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
```
rushtong: Hmmm ... this file should be in `.gitignore`. Can you revert it for this PR? I'll look into why it's showing up at all.

src/main/webapp/search/SearchResults.js
```diff
@@ -1,136 +1,231 @@
...
+      if (aIsNewer) return 1;
+      if (bothSame) return 0;
+      return -1;
     }
-}
+  };
 
-function expirationSort(a, b, order) {
+  expirationSort = (a, b, order) => {
     return compareDateString(a.expiration, b.expiration, order);
```
rushtong: Since this is no longer a function call, should this be `this. compareDateString` ? Both here and below.

rushtong:
Also - I think this needs a merge from latest feature - the compiled search.js file should be ignored.

grails-app/domain/org/broadinstitute/orsp/Issue.groovy
LeoForconesi: @rushtong @dags  
`getExtraPropertiesMap()`'s goal is to obtain some sort of a "maped" object {key: value} style to be send directly to the Ui, and save some parse methods in this end. 
I belive that if we now have a `getCollaborators` method, maybe we could use it thogether with `getExtraPropertiesMap` ?. 
Our Ui expects a "Collaborators" extraProperty which contains an array of collaborators string names. Here's a [link](https://github.com/broadinstitute/orsp-pub/blob/feature/src/main/webapp/projectReview/ProjectReview.js#L129) to see how it's used on the front end. 
Does it clarify your doubts? Let me know

rushtong: Thanks @LeoForconesi. I spent a little more time looking at that and I'll submit a follow-up PR that I think will clean it up. The essence of that method is to copy all `collaborator` properties into a `collaborators` map entry.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -0,0 +1,40 @@
...
+      issue.extraProperties.findAll ({ it.name == IssueExtraProperty.PM }).collect { property -> property.value }
+    }
+
+    // get issue's pis as a List<String>
+    List<String> getIssuePIs(Issue issue) {
+      issue.extraProperties.findAll ({ it.name == IssueExtraProperty.PI }).collect { property -> property.value }
+    }
+
+    // verifies if logged user belongs to some user list ....
+    def issueIsForbidden(issue, userName, isAdmin, isReadOnlyAdmin) {
```
rushtong: These should also be typed:
```suggestion
    def issueIsForbidden(Issue issue, String userName, boolean isAdmin, boolean isReadOnlyAdmin) {
```

PR #97, opened 2019-03-15
[risk=no] Another Refactor

rushtong:
## Addresses
No ticket

## Changes
Minor refactoring

---

### DIFF
--- a/grails-app/domain/org/broadinstitute/orsp/Issue.groovy
+++ b/grails-app/domain/org/broadinstitute/orsp/Issue.groovy
@@ -195,22 +195,22 @@ class Issue implements LogicalDelete<Issue> {
         properties
     }
 
-    transient LinkedHashMap<String, Object> getExtraPropertiesMap() {
-        LinkedHashMap<String, Object> extraproperties = new LinkedHashMap<>()
-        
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/broadinstitute/orsp/Issue.groovy
```diff
@@ -195,22 +195,22 @@ class Issue implements LogicalDelete<Issue> {
...
-                extraproperties.put(it.name, it.value)
-            } else {
-              collaborators.add(it.value);
-            }
+    /**
+     * Returns a map of IssueExtraProperty values for an issue, grouped by name.
+     *
+     * @return Map of extra properties, grouped by name
+     */
+    @SuppressWarnings("GroovyAssignabilityCheck")
```
rushtong: This is needed to stop the IDE from complaining about `propsByName.collectEntries`. It should know that `key` is a string and `list` is a `List<String>`, but doesn't.

LeoForconesi:
Ok, I've tested in my local and even play around with debug, pushing this method, and it seems to work.

rushtong:
Thank you @LeoForconesi !

PR #98, opened 2019-03-19
BTRX - Extra properties fix

LeoForconesi:
## Changes
- Created two object classes for Project Extra Properties and for Consent Group Extra Properties
- Each class, has a builder that receives a Map of extra properties and returns the specific object type
- Changed some default values for the Ui

---


src/main/groovy/org/broadinstitute/orsp/ConsentGroupExtraProperties.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+    // InternationalCohorts
+    String individualDataSourced = ""
+    String isLinkMaintained = ""
+    String feeForServiceWork = ""
+    String areSamplesComingFromEEAA = ""
+    String isCollaboratorProvidingGoodService = ""
+    String isConsentUnambiguous = ""
+
+    transient ConsentGroupExtraProperties buildConsentGroupExtraProperties(Map<String, List<String>> extraPropsMap) {
```
rushtong: I don't think you need `transient` here - that's used to indicate transience on domain objects that require saving to the db.

dags (approved):

src/main/groovy/org/broadinstitute/orsp/ConsentGroupExtraProperties.groovy
```diff
@@ -0,0 +1,81 @@
...
+        if(!extraPropsMap.isEmpty()) {
+            // General Data
+            this.setStartDate(extraPropsMap.get(IssueExtraProperty.START_DATE, [""]).first())
```
rushtong: My concern here is that there is a ton of duplicated code (both here and in the `ProjectExtraProperties` class). We already do this kind of thing in [`Issue`](https://github.com/broadinstitute/orsp-pub/blob/bb6507f310a87f1ba4875f6fbee001bca26cfb5c/grails-app/domain/org/broadinstitute/orsp/Issue.groovy#L69) and, to some degree, it makes more sense for all of this code to be using that mechanism instead of replicating this logic here.

rushtong: I think it would be simpler to make the constructor take an `Issue` object and then populate the values from what is already known on the issue itself.

rushtong (changes requested):
I think you're on the right track, but I'd like to see the extra prop logic happen in one consistent place. Right now, that's on `Issue` which is where I think it should stay.

LeoForconesi:
@rushtong  this is ready for your review

rushtong (approved):
Thank you for the changes!

rushtong (approved):
Still 👍

PR #99, opened 2019-03-25
Btrx 581 send email project trk

tadeork:
## Addresses
ORSP: Send email after Project creation based on new tabs answers.
https://broadinstitute.atlassian.net/browse/BTRX-581
## Changes
The notifications are the same for CG. I've re used some methods and added new validations.
## Testing

---


grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
```diff
@@ -499,14 +501,17 @@ class NotifyService implements SendgridSupport, Status {
...
-    Map<Boolean, String> sendConsentGroupSecurityInfo(Issue issue, User user) {
+    Map<Boolean, String> sendRequirementsInfo(Issue issue, User user, String type) {
         Map<String, String> values = new HashMap<>()
         Map<Boolean, String> result = new HashMap<>()
 
         if (Boolean.valueOf(issue.getMTA())) {
             values.put(IssueExtraProperty.REQUIRE_MTA, "true")
         }
-        if (issue.getFeeForService() != null && Boolean.valueOf(issue.getFeeForService())) {
+        if (type != 'Project' && issue.getFeeForService() != null && Boolean.valueOf(issue.getFeeForService())) {
```
rushtong: `type` here really should be an enum. I try not to do raw string comparisons. I don't think we have an enum that exactly matches what you want here, but it seems to me that the best place for this would be `IssueType`.

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
```diff
@@ -574,13 +581,15 @@ class NotifyService implements SendgridSupport, Status {
     }
 
     def projectCGCreation(Issue issue) {
```
rushtong: I must have missed this in an earlier review - this should be returning `Map<Boolean, String>`

grails-app/views/notify/requirements.gsp
```diff
@@ -25,11 +25,16 @@
         </ul>
     </div>
 <p>
-    If you have any questions or concerns, please contact our colleague who created this Consent Group and is cc’d on this email.
+    A member from the OSAP team, cc’d here will be in touch with you shortly to proceed with obtaining them.
 </p>
+<g:if test="${values.get("type").equals("Project") || values.get("feeForService")?.equals("true") || values.get("areSamplesComingFromEEAA")?.equals("true") || values.get("isConsentUnambiguous")?.equals("true")}">
```
rushtong: `"Project"` here should also be an enum for safer refactoring/typing. Here is one way to do that: https://www.sohamkamani.com/blog/2017/08/21/enums-in-javascript/#enumerations-with-objects

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
tadeork: Could we solve this using a constant inside the class? IssueType is used to filter documents by type of issue. If I add another value the logic that uses IssueType it will try to filter by this new value.

rushtong: Yes, that could also work. Yes, `IssueType` might not be the best place for this, but it's where I would *expect* something like `Project` type.

grails-app/views/notify/requirements.gsp
tadeork: This case is a template to render the email content. I don't think that is good idea to add js to this. Maybe the best idea is to have an enum with two values Project and CG and use it here and the previous case.

tadeork: No. I tried to solve this with an enum in the template but it didn't work. Can we merge this any way?

PR #104, opened 2019-03-27
[BTRX-523][risk=no] Use single list of attachment types across different projects (#57)

rushtong:
## Addresses
Addresses https://broadinstitute.atlassian.net/browse/BTRX-523
Copy of https://github.com/broadinstitute/orsp-pub/pull/57 against the `feature` branch

## Changes
See https://github.com/broadinstitute/orsp-pub/pull/57

---


rushtong:
@wloforteBroad - This is a very specific merge from master that should only affect the older modals to add a list of document types. I am more concerned that this doesn't affect the new workflows - which I did some testing in, but was not extensive, so I do appreciate your help in testing that out.

wloforteBroad:
I will, as soon as I can and will let you know.

PR #105, opened 2019-03-27
Add Info Security to Project Review and Edit 

wloforteBroad:
On behalf of @tadeork 

**Addresses:**
(https://broadinstitute.atlassian.net/browse/BTRX-586)

The Security questions already added to the creation flow are now on the Review Page and are editable.

---


src/main/webapp/components/SecurityReview.js
```diff
@@ -0,0 +1,219 @@
...
+        accessible: true,
+        textCompliance: true,
+        textSensitive: true,
+        textAccessible: true,
+      },
+    };
+    this.formHasError = this.formHasError.bind(this);
+  }
+
+  componentDidMount() { }
```
rushtong: Do you need this?

src/main/webapp/components/SecurityReview.js
```diff
@@ -0,0 +1,219 @@
...
+        edit: this.props.edit,
+        review: this.props.review
+      })
+    ])
+  };
+
+  render() {
+    if (this.state.hasError) {
+      // You can render any custom fallback UI
+      return h1({}, ["Something went wrong."]);
```
rushtong: This seems pretty unhelpful. Can we add something explanatory here?

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -376,6 +376,7 @@ class NewConsentGroup extends Component {
   };
 
   handleInfoSecurityValidity(isValid) {
+    console.log('handleInfoSecurityValidity', isValid);
```
rushtong: rm

src/main/webapp/components/SecurityReview.js
tadeork: I agree. This should be more explanatory. But we're using this as a simple way to display a error when the component can't load. Probably we need to fix this later to a better explanation, in all the components this messages is present. At least for us this message it's helpful for now.

PR #106, opened 2019-03-27
Refactor the react index.gsp files

rushtong:
## Changes
* Make all of the primary react-based `index.gsp` files use the same template. 
* Pull all constants into the same object(s) so they can be reused.

## Testing
View search, project, consent group pages - should be no errors

## TODO
There is a lot we can do to refactor further, but I leave that for other PRs. This PR is meant to start us in the right direction in terms of consolidating react behavior. The more we can pull out of individual gsps, the easier it will be to consolidate them later on as well as convert other features more easily.

Also - this does not address the templates we are using (`_index.gsp` files). We need to figure that out too. 

---

### DIFF
--- a/grails-app/views/layouts/main_component.gsp
+++ b/grails-app/views/layouts/main_component.gsp
@@ -1,0 +1,184 @@
+<%@ page import="org.broadinstitute.orsp.IssueStatus" %>
+<%@ page import="org.broadinstitute.orsp.IssueType" %>
+<%@ page import="org.broadinstitute.orsp.PreferredIrb" %>
+
+<!DOCTYPE html>
+<html>
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+    <title><g:layoutTitle default="ORSP Portal"/></title>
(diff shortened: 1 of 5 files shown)


grails-app/views/layouts/main_component.gsp
```diff
@@ -0,0 +1,184 @@
+<%@ page import="org.broadinstitute.orsp.IssueStatus" %>
+<%@ page import="org.broadinstitute.orsp.IssueType" %>
+<%@ page import="org.broadinstitute.orsp.PreferredIrb" %>
+
+<!DOCTYPE html>
```
rushtong: This file is a straight copy of `main.gsp` with additions to support the react stuff. I envision this being where we start doing the consolidation and eventually deprecate `main.gsp` altogether.

tadeork: So in the future we'll use this as a main entry-point to all the info that the components will need?

rushtong: That is the idea, yes. We could instead use the old `main.gsp` but I wanted to try and minimize the introduction of bugs with the old code.

rushtong:
Thanks @tadeork - will merge and if we see problems, will address immediately.

PR #110, opened 2019-03-28
Btrx 587 int cohorts

LeoForconesi:
## Addresses
[BTRX-587](https://broadinstitute.atlassian.net/browse/BTRX-587): ORSP: Add International Cohorts Questionnaire to NE and NHSR Projects Review and Edit flow

## Changes
- Due to project extra properties already contains a variable named `feeForService` we used `feeForServiceWork` name to differenciate one and another. In order for this change to be updated on  'issue' object, we had to map this variable in the backend

- Created a new component for International Cohorts review so we save boilerplate code in Project and Consent Group

- Some minor tweaks had to be done on `questionnaireWorkflow`. Essentially to differentiate when this questionnaire is being used for creation or edition

- Keep centralizing the use of `isEmpty` method placed in Utils

---

### DIFF
--- a/src/main/webapp/consentGroupReview/ConsentGroupReview.js
+++ b/src/main/webapp/consentGroupReview/ConsentGroupReview.js
@@ -12,17 +12,11 @@ import { ConsentGroup, SampleCollections, User, Review } from "../util/ajax";
 import { RequestClarificationDialog } from "../components/RequestClarificationDialog";
 import { ConfirmationDialog } from "../components/ConfirmationDialog";
 import { spinnerService } from "../util/spinner-service";
-import { QuestionnaireWorkflow } from "../components/QuestionnaireWorkflow";
 import { AlertMessage } from "../components/AlertMessage";
 import { Spinner } from "../components/Spinner";
 import get from 'lodash/get';
 import { format } from 'date-fns';
-
-const EXIT = 500;
-const DPA = 600;
-const RA = 700;
-const CTC = 800;
-const OSAP = 900;
+import { IntCohortsReview } from "../components/IntCohortsReview";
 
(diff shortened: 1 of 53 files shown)


src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -579,12 +579,30 @@ class ConsentGroupReview extends Component {
...
+    this.setState(prev => {
+      prev.resetIntCohorts = false;
+      prev.determination = this.resetIntCohortsDetermination();
+      return prev;
+    })
+  }
+
+  resetIntCohortsDetermination() {
+    return {
+      projectType: 900,
```
rushtong: Is there a list of what these numbers correspond to? This would be very confusing to a dev looking at this unless they knew the context (like me ... and I don't know the context). If the values can be constants defined somewhere with descriptive names, that would be more helpful.

LeoForconesi: In order to prevent any confusion, I created an enum(ish) object that contains all possible determination types outputs coming from questionnaire workflow implementation. So now they are all together in one place. 

On the other hand, it wans't neccesary to inicialize any type of determination for this object, you will see them null on constructors and for reset purpouses.

LeoForconesi:
@rushtong i Corrected the code, please check again. 
Also added a validation scenario for Legacy projects. In these cases International Cohorts and Info Security shouldn't be required to submit an edit/suggestion

src/main/webapp/consentGroupReview/ConsentGroupReview.js
rushtong: 👍

PR #111, opened 2019-03-28
BTRX - 602 & 591  - Put 'Add New Consent Group' button in Project review - Project type Api

LeoForconesi:
## Addresses
[BTRX-602](https://broadinstitute.atlassian.net/browse/BTRX-602): ORSP - Project Info tab: Put the new 'Add New Consent Group' button next to the 'Edit' button
[BTRX-591](https://broadinstitute.atlassian.net/browse/BTRX-591): ORSP - Technical debt - Create API to get project type

## Changes
- Code related to mapping `consentUpload` new variable (creation and update)
- Created a new backend api that returns a project type, given a projectKey. This was a technical debt.

##
**Note (Just as a discussion starter):**
 It would be really interesting to correct webpack.config in order to enable the use of async/await functionallity in our code. This would make promises handling easier and shorten our code. 

Apparently the problem is that `regeneratorRuntime` is undefined, this is due to  having a custom webpack.config where babel-polyfill is not being import. I tried to do it quickly here, but it demands a little more investigation. 

---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/IssueService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/IssueService.groovy
@@ -447,15 +447,8 @@ class IssueService {
     }
 
     String getProjectType(String projectKey) {
-        String projectType
         Issue issue = queryService.findByKey(projectKey)
-        if(issue != null) {
-            projectType = IssueType.valueOfName(issue.getType()).prefix.toLowerCase()
-            projectType
-        } else {
-            projectType = null
-        }
-        projectType
+        IssueType.valueOfName(issue?.getType())?.prefix?.toLowerCase()
     }
 
 }
--- a/grails-app/services/org/broadinstitute/orsp/IssueService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/IssueService.groovy
@@ -431,4 +431,16 @@ class IssueService {
         props
     }
 
+    String getProjectType(String projectKey) {
+        String projectType
+        Issue issue = queryService.findByKey(projectKey)
+        if(issue != null) {
+            projectType = IssueType.valueOfName(issue.getType()).prefix.toLowerCase()
+            projectType
+        } else {
+            projectType = null
+        }
+        projectType
+    }
+
 }
--- a/src/main/webapp/consentGroup/NewConsentGroup.js
+++ b/src/main/webapp/consentGroup/NewConsentGroup.js
@@ -4,7 +4,7 @@ import { NewConsentGroupDocuments } from './NewConsentGroupDocuments';
 import { NewConsentGroupGeneralData } from './NewConsentGroupGeneralData';
 import { InternationalCohorts } from '../components/InternationalCohorts';
 import { span, a } from 'react-hyperscript-helpers';
-import { Files, ConsentGroup, SampleCollections, User } from '../util/ajax';
+import { Files, ConsentGroup, SampleCollections, User, Project } from '../util/ajax';
 import { spinnerService } from '../util/spinner-service';
 import { DataSharing } from '../components/DataSharing';
 import { Security } from '../components/Security';
@@ -134,7 +134,11 @@ class NewConsentGroup extends Component {
   uploadFiles = (projectKey) => {
     Files.upload(this.props.attachDocumentsURL, this.state.files, projectKey, this.state.user.displayName, this.state.user.userName, true)
       .then(resp => {
-        window.location.href = this.getRedirectUrl();
+
+        Project.getProjectType(this.props.serverURL, this.props.projectKey).then(type => {
+          window.location.href =  [this.props.serverURL, type.data.projectType, "show", this.props.projectKey, "?tab=consent-groups"].join("/");
+        });
+        // window.location.href = this.getRedirectUrl();
         spinnerService.hideAll();
         this.setState(prev => {
           prev.formSubmitted = true;
@@ -226,10 +230,6 @@ class NewConsentGroup extends Component {
     }
     return renderSubmit;
   };
-
-  getRedirectUrl() {
-    return [this.props.serverURL, this.props.projectType, "show", this.props.projectKey, "?tab=consent-groups"].join("/");
-  }
 
   isValid = (field) => {
     let isValid = true;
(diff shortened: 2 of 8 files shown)


grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -446,4 +446,16 @@ class IssueService {
         props
     }
 
+    String getProjectType(String projectKey) {
+        String projectType
+        Issue issue = queryService.findByKey(projectKey)
+        if(issue != null) {
```
rushtong: nit, spacing
```suggestion
        if (issue != null) {
```

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -446,4 +446,16 @@ class IssueService {
...
+            projectType = IssueType.valueOfName(issue.getType()).prefix.toLowerCase()
+            projectType
+        } else {
+            projectType = null
+        }
+        projectType
```
rushtong: This could be simplified with the null-safe operator. Something like this
```
String type = issue?.getType()
projectType = Optional.ofNullable(type).getOrElse("")
```
Which could be shortened to a one-liner if needed.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -134,7 +134,9 @@ class NewConsentGroup extends Component {
   uploadFiles = (projectKey) => {
     Files.upload(this.props.attachDocumentsURL, this.state.files, projectKey, this.state.user.displayName, this.state.user.userName, true)
       .then(resp => {
-        window.location.href = this.getRedirectUrl();
+        Project.getProjectType(this.props.serverURL, this.props.projectKey).then(type => {
+          window.location.href =  [this.props.serverURL, type.data.projectType, "show", this.props.projectKey, "?tab=consent-groups"].join("/");
+        });
```
rushtong: Ouch ... I'm pretty sure I missed this in previous reviews. We should not be reloading the page in react components. These should trigger a react lifecycle event instead if we think we need to reload page data. State should drive what happens in the view here. Ideally, we'd fire off an async post, then update state with the results.

src/main/webapp/project/NewProject.js
```diff
@@ -509,7 +509,9 @@ class NewProject extends Component {
   uploadFiles = (projectKey) => {
     Files.upload(this.props.attachDocumentsURL, this.state.files, projectKey, this.state.user.displayName, this.state.user.userName, true)
       .then(resp => {
-        window.location.href = this.getRedirectUrl(projectKey);
+        Project.getProjectType(this.props.serverURL, projectKey).then(type => {
+          window.location.href =  [this.props.serverURL, type.data.projectType, "show", projectKey, "?tab=review"].join("/");
+        });
```
rushtong: See previous comment about state.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -703,6 +703,9 @@ class ProjectReview extends Component {
       return prev;
     });
   };
+  redirectToNewConsentGroup = () => {
+    window.location.href = this.props.serverURL + '/api/consent-group?projectKey=' + this.props.projectKey;
+  };
```
rushtong: Here's another one - I would like to rethink this pattern we're using.

rushtong:
@LeoForconesi @tadeork 
If we change `.babelrc` to:
```
{
  "presets": [
    ["env", {
      "targets": {
        "browsers": [
          ">0.25%",
          "not ie 11",
          "not op_mini all"
        ]
      }
    }], "react", ["es2015"]
  ]
}
```
Does that work for you? I found [this link](https://www.valentinog.com/blog/how-async-await-in-react/) which explains the specific issue mentioned.

src/main/webapp/consentGroup/NewConsentGroup.js
LeoForconesi: I understand your concern here, but this method only intends to redirect to preview page after creation, it's not reloading the same page.

On the other hand, it's  prefered to use React Routes, which could be used after we have a Higher order Component making the link between Grails and React, and containing a Switch to redirect each path with its props.

Either way i'll make a change here adding Async/Await in order to prevent this redirection inside a promise.

LeoForconesi:
> @LeoForconesi @tadeork
> If we change `.babelrc` to:
> 
> ```
> {
>   "presets": [
>     ["env", {
>       "targets": {
>         "browsers": [
>           ">0.25%",
>           "not ie 11",
>           "not op_mini all"
>         ]
>       }
>     }], "react", ["es2015"]
>   ]
> }
> ```
> Does that work for you? I found [this link](https://www.valentinog.com/blog/how-async-await-in-react/) which explains the specific issue mentioned.
> 
> **UPDATE**: I tested this locally by adding an async on a search page method and it doesn't work.
> **UPDATE**: OK, I have a preliminary fix. Add `"regenerator-runtime": "0.13.2"` to `package.json` and then you can `import "regenerator-runtime/runtime";` in your code where you need `async`. Ideally, we'd just run the import once but since we have separate react files, we need to import that where needed. In the long run, we can roll this up into the an across the board place.

This Works!

rushtong:
@LeoForconesi - I don't see any other changes since my previous review. Let me know if there's anything new to look at here. Thank you!

src/main/webapp/consentGroup/NewConsentGroup.js
rushtong: OK, I see what you're trying to do here. The effect of `window.location.href` is to reload the page with the destination. In react apps, we should never have to set the window's location. In our world, we are transitioning from older style gsps, so we might need to redirect, but we absolutely need to make sure this is temporary code. 

We need two things here, a comment explaining what we're doing and a link to a tech-debt ticket that covers fixing this. With those two things, 👍

LeoForconesi: Here's the tech-debt story.
[BTRX-637](https://broadinstitute.atlassian.net/browse/BTRX-637) : ORSP - Technical debt - improve react redirections

LeoForconesi:
@rushtong 
Could please check now?

PR #112, opened 2019-03-29
BTRX-627 ORSP - Several text changes

nlopezzalba:
## Addresses

Page/s: Consent Group creation and review
Update the first International Cohorts question to include the link: “Are samples or individual-level data sourced from a country in the European Economic Area? https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm”

Page: Project/Consent Group creation
Add the following text below the main page title on both Project and Consent Group creation workflows: “Note that this application cannot be saved and returned to for completion later. However, allowing the page to remain open in your browser will permit you to return to the application at any time.”

Pages: Project review > Request Clarification dialog
Change text on Clarifications popup box button to say ‘Request Clarification’ instead of ‘Submit Clarification’

Page/s: Project creation and review
Update the text of the Determination Question (see attached image below) to: “Are samples/data being provided by an investigator who has access to identifiers or obtains samples through an intervention or interaction?”

Page/s: Project/Consent Group creation and review
In the second question on the Security tab, remove “CLIA” and write “HIPAA”

Page/s: Project creation and review
Remove ‘Prime’ before ‘Sponsor Name” in the Funding box of the General Data/Project Details tab

## Testing

---

### DIFF
--- a/src/main/webapp/components/InternationalCohorts.js
+++ b/src/main/webapp/components/InternationalCohorts.js
@@ -34,7 +34,7 @@ export const InternationalCohorts = hh(class InternationalCohorts extends Compon
     questions.push({
       question: span({}, [
         "Are samples or individual-level data sourced from a country in the European Economic Area? ", 
-        a({ href:"https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm", target: "_blank", className: "normal", style: { 'display': 'inline' } }, ["[https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm]"])
+        a({ href:"https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm", target: "_blank", className: "normal" }, ["[https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm]"])
     ]),
       yesOutput: 2,
       noOutput: EXIT,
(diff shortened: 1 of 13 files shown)


src/main/webapp/components/InternationalCohorts.js
```diff
@@ -32,7 +32,10 @@ export const InternationalCohorts = hh(class InternationalCohorts extends Compon
     let questions = [];
 
     questions.push({
-      question: span({}, ["Are samples or individual-level data sourced from a country in the European Economic Area? ", span({ className: "normal" }, ["[provide link to list of countries included]"])]),
+      question: span({}, [
+        "Are samples or individual-level data sourced from a country in the European Economic Area? ", 
+        a({ href:"https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm", target: "_blank", className: "normal" }, ["(List of member states of European Economic Area)"])
```
rushtong: nit: I think this string array can just be a string (here and in other places too). Looking at the [documentation](https://www.npmjs.com/package/react-hyperscript-helpers), this can be an array, boolean, string, number, etc. I do know that when I tried that with a `div`, it threw an error (or I messed up ... very likely) and expected an array, so I think the docs might be a little fuzzy on that detail.
```suggestion
        a({ href:"https://www.imf.org/external/pubs/ft/fandd/2014/03/europeaneconomicarea.htm", target: "_blank", className: "normal" }, "(List of member states of European Economic Area)")
```

nlopezzalba: oh nice! didn't know that. I'll change it, thanks!

PR #116, opened 2019-04-04
BTRX-595 ORSP - New Project/Consent Group Generation Workflow Add Additional Documents

vvicario:
## Testing

---


src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -537,7 +479,7 @@ class NewConsentGroup extends Component {
     }).catch(error => {
       console.error(error);
     });
-  };
+  }
```
rushtong: I see a lot of this kind of churn where one person on the team adds semi-colons and others remove them. Can we look into a standard? Maybe a lint file that enforces style would be helpful.

src/main/webapp/consentGroup/NewConsentGroupDocuments.js
```diff
@@ -1,14 +1,39 @@
...
+const addDocumentBtn = {
+  position: 'relative', float: 'right'
+};
+
+const headers =
+  [
+    { name: 'Document Type', value: 'fileKey' },
+    { name: 'File Name', value: 'fileName' },
+    { name: '', value: 'remove' }
+  ];
```
rushtong: nit: Since these are all styles, it might make sense to have one const for styles and then have each one of these three const values as an object inside. That would make the code more clearly self-documenting.

rushtong: Oops, this comment applies to `addDocumentContainer` and `addDocumentBtn`, not `headers`, sorry.

src/main/webapp/consentGroup/NewConsentGroup.js
vvicario: sure!

vvicario:
Agree!

PR #132, opened 2019-04-18
[BTRX-654] Update eDUL PDF

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-654
## Changes
Removed fields of diseases and added new field for multiple diseases selection. It's related to
https://broadinstitute.atlassian.net/browse/BTRX-640
## Testing

---


src/main/groovy/org/broadinstitute/orsp/dataUseLetter/DataUseLetterFields.groovy
```diff
@@ -20,24 +20,13 @@ enum DataUseLetterFields {
...
-    SKIN_DISEASE("skinDisease"),
-    MUSCULO_SKELETAL_DISEASE("musculoskeletalDisease"),
-    GENITOURINARY_DISEASE("genitourinaryDisease"),
     CARDIOVASCULAR_DISEASE("cardiovascularDisease"),
-    PREGNANCY("pregnancy"),
-    CONGENITAL_MALFORMATION("congenitalMalformation"),
-    BLOOD_DISORDER("bloodDisorder"),
     OTHER_DISEASE("otherDisease"),
-    OTHER_DISEASE_SPECIFY("otherDiseaseSpecify"),
+    OTHER_DISEASE_DOID("otherDiseasesDOID"),
```
rushtong: Nit: naming. This naming suggests (not incorrectly) that the disease ID will be a DOID. This is a tight coupling to the ontology. While it is true that we are using DOID today, we may not be using DOID in the future. I think OTHER_DISEASE_ID is just as descriptive and yet does not preclude other ontology disease ids.

PR #134, opened 2019-04-25
BTRX-665-Move Attestation question

wloforteBroad:
Addresses:

Move the Attestation question from General Data Tab to last Step before Submit.

https://broadinstitute.atlassian.net/browse/BTRX-665

---

### DIFF
--- a/src/main/webapp/project/NewProjectDocuments.js
+++ b/src/main/webapp/project/NewProjectDocuments.js
@@ -145,18 +145,20 @@ export const NewProjectDocuments = hh(class NewProjectDocuments extends Componen
               remove: this.removeFile,
               reviewFlow: false
             }),
-            Panel({ title: "Broad Responsible Party (or Designee) Attestation*" }, [
-              p({}, 'I confirm that the information provided above is accurate and complete. The Broad researcher associated with the project is aware of this application, and I have the authority to submit it on his/her behalf.'),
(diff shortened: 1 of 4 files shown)


src/main/webapp/project/NewProjectDocuments.js
```diff
@@ -1,7 +1,8 @@
 import { Component, Fragment } from 'react';
 import { WizardStep } from '../components/WizardStep';
-import { hh, h, h1, span, button, div } from 'react-hyperscript-helpers';
+import { hh, h, h1, span, button, div, p, small } from 'react-hyperscript-helpers';
```
rushtong: Not related to this PR, but I'd like folks' opinions on if `react-hyperscript-helpers` has been a time saver or not ... I'm sort of on the fence about this. Will ask in standup.

src/main/webapp/project/NewProjectDocuments.js
```diff
@@ -75,6 +82,16 @@ export const NewProjectDocuments = hh(class NewProjectDocuments extends Componen
     });
   };
 
+  handleAttestationCheck = (e) => {
+    let value = e.target.checked;
```
rushtong: this could be a const ... it's not getting modified anywhere.

wloforteBroad: You are totally right. I'll change this.

PR #136, opened 2019-04-29
BTRX-657 eDUL form submission generates sDULs

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-657
## Changes
Move restriction logic to data use letter service
## Testing

---


src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -553,9 +557,73 @@ class DataUseLetter extends Component {
...
+  }
+
+  getDiseases(diseases) {
+    let diseasesResult = [];
+    diseases.map(disease => {
+      diseasesResult.push(
+        disease.key
+      );
+     });
+    return diseasesResult
```
rushtong: Pretty sure you don't need the interim `diseasesResult` variable ... the return value of `map` is an array populted with the results of whatever happened inside of the map.

src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -553,9 +557,73 @@ class DataUseLetter extends Component {
...
+  }
+
+  getRestriction() {
+    let diseaseRestrictions = [];
+    let hasDiseases = false;
+    if (this.state.formData.diseaseRestrictedOptions.diseaseDOID.length > 0) {
```
rushtong: Do you need to protect against any empty data here? This looks brittle.

src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -553,9 +557,73 @@ class DataUseLetter extends Component {
...
+      diseaseRestrictions = [...this.getDiseases(this.state.formData.diseaseRestrictedOptions.diseaseDOID)];
+      hasDiseases = true;
+    }
+    if (this.state.formData.otherDiseasesID.length > 0) {
```
rushtong: similar concern here.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseRestrictionDTO.groovy
```diff
@@ -0,0 +1,94 @@
+package org.broadinstitute.orsp.consent
+
+import org.broadinstitute.orsp.utils.IssueUtils
+import org.grails.web.json.JSONArray
+
+class DataUseRestrictionDTO {
```
rushtong: I'm not sure I understand the need for a DTO here. Why not just put the `fromParams` method into a utility class and return a new `DataUseRestriction` ? There's a ton of copy/paste going on between this class, `DataUseRestriction`, and also in the `DataUserLetterService.createSdul` method. Having to maintain all of these duplicate fields in three places is going to be painful over time. 

Note that even though you can't populate a complete DUR from the input, the vast majority of it *can* be populated. What you'd really be creating with a utility method is a partial DUR. You can finish the population in the `createSdul` method.

rushtong:
Sorry, meant to add a comment to the summary. Thanks for the update - it's definitely better, but I think we can make this a lot simpler and reduce the amount of duplicate code.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseRestrictionDTO.groovy
vvicario: Agree! I worked on that

vvicario:
@rushtong I worked on your feedback. Please let me know if I need to change something else

PR #137, opened 2019-05-02
BTRX-669-Change Field in Admin Tab

wloforteBroad:
## Addresses

https://broadinstitute.atlassian.net/browse/BTRX-669

Admin Tab: Please add the Drop-down feature under "Type of Initial Review" (current functionality in production under Review Category) Expedited, Exempt, Full Board, Not Human Subjects Research, Not Engaged* (*new choice).

---

### DIFF
--- a/src/main/webapp/adminOnly/AdminOnly.js
+++ b/src/main/webapp/adminOnly/AdminOnly.js
@@ -10,6 +10,7 @@ import { format } from 'date-fns';
 import "regenerator-runtime/runtime";
 import { InputFieldSelect } from "../components/InputFieldSelect";
 import { PREFERRED_IRB } from "../util/TypeDescription";
+import { INITIAL_REVIEW } from "../util/TypeDescription";
 import { InputTextList } from "../components/InputTextList";
 import { Fundings } from "../components/Fundings";
 import { Spinner } from "../components/Spinner";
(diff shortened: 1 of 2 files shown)


src/main/webapp/adminOnly/AdminOnly.js
```diff
@@ -184,7 +185,7 @@ class AdminOnly extends Component {
     form.investigatorFirstName = this.state.formData.investigatorFirstName;
     form.investigatorLastName = this.state.formData.investigatorLastName;
     form.initialDate = this.parseDate(this.state.formData.initialDate);
-    form.initialReviewType = this.state.formData.initialReviewType;
+    form.initialReviewType = JSON.stringify(this.state.formData.initialReviewType);
```
rushtong: Curious about why this needs to be stringified - usually that's for display purposes. I'm pretty sure it doesn't do anything structurally to the data here.

dags: Select component returns selected value as an json object { label:"xxxx", value: "zzz"}. If we store only selectedValue.value then we need to recreate selected Value at read. We saved as stringified, we parse at read.

wloforteBroad: @rushtong can we merge this PR?

rushtong: Absolutely ... I approved it!

PR #150, opened 2019-05-10
BTRX-675 Implement Add New Sample/Data Cohort wizard

vvicario:
## Testing

---


vvicario:
Hi Greg:

1) We decided to add project key and consent key to ConsentCollectionData in this way we can associate with ConsentCollectionLink where we have sample collections
2) As far as we understand, we can't associate documents to ConsentCollectionData (link). Documents can be associated in: The Documents tab in a New Project, 2.  the Documents tab in a Sample/Data Cohort, and 3. In the Submission Tab.
3) It's not necessary to migrate ConsentCollectionLink we want to keep this information in this table, we are going to have only one row in ConsentCollectionData associated to multiple rows in ConsentCollectionLink, depends on the amount of consent collections associated to consent and project

rushtong:
@dags, @vvicario - We need to discuss this technical design. I'm pretty sure there are some incorrect assumptions being made here. Without having a detailed description of what you're doing, it's hard for me to understand how this will work.

vvicario:
@rushtong please can you take a look? Let us know if you have any questions. Thanks!

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -45,51 +46,34 @@ class NewConsentGroupController extends AuthenticatedController {
...
+        Issue consent
+        ConsentCollectionLink consentCollectionLink
         try{
             List<MultipartFile> files = request.multiFileMap.collect { it.value }.flatten()
             User user = getUser()
-            String dataProject = request.parameterMap["dataProject"].toString()
             JsonParser parser = new JsonParser()
-            JsonArray dataProjectJson = parser.parse(dataProject)
+            JsonArray dataProjectJson = parser.parse(request.parameterMap["dataProject"].toString())
+            JsonArray dataConsentCollectionJson = parser.parse(request.parameterMap["dataConsentCollection"].toString())
```
rushtong: nit: not necessarily for this PR, but I really wish we had a better way of doing this, or at least packaging it into a utility class so we're not doing so much work in the controller. This pattern of pulling out a json array, then coercing into a domain object is very brittle. There is no error handling here which means if something is not correct in the `request.parameterMap`, the whole thing will fail with just a flash error message. 

It might make more sense to:
* compose the various objects from the request params
* inspect each of them for validity
* return a more useful error in the case that anything is not correct
* perform the save operation on the validated objects.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -45,51 +46,34 @@ class NewConsentGroupController extends AuthenticatedController {
...
-                    }
+                    consentCollectionLink.save(flush: true)
                 } catch (Exception e) {
                     flash.error = e.getMessage()
                 }
-
                 if (!files?.isEmpty()) {
                     files.forEach {
-                        storageProviderService.saveMultipartFile(user.displayName, user.userName, consent.getProjectKey().toString(), it.name, it)
+                        storageProviderService.saveMultipartFile(user?.displayName, user?.userName, issue.getSource(), it.name, it, issue.projectKey)
```
rushtong: If the user displayname and username are null, shouldn't that be an error?

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -163,7 +163,7 @@ class StorageProviderService implements Status {
      * @param files The multipart files
      * @return
      */
-    StorageDocument saveMultipartFile(String displayName, String userName, String issueKey, String type, MultipartFile file) {
+    StorageDocument saveMultipartFile(String displayName, String userName, String issueKey, String type, MultipartFile file, String consentKey) {
```
rushtong: This is a mistake, please revert. `issueKey` is where you should be storing the key of the consent's project key. This will end up breaking all existing storage documents for consent groups where the project key of the consent group is already stored in `issueKey`.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -369,7 +431,8 @@ class NewConsentGroup extends Component {
 
   validateInfoSecurity() {
     this.setState(prev => {
-      prev.showInfoSecurityError = !this.state.isInfoSecurityValid;
+      prev.showErrorInfoSecurity = !this.state.isInfoSecurityValid;
+
```
rushtong: nit: why the white space change?

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -444,62 +518,52 @@ class NewConsentGroup extends Component {
...
-          submitError: this.state.submitError,
-          removeErrorMessage: this.removeErrorMessage,
-          handleSecurityValidity: this.handleInfoSecurityValidity,
-          currentValue: this.state,
-          edit: false,
-          review: false,
-          readOnly: false
-        })
-      ])
+          NewConsentGroupGeneralData({
```
rushtong: nit: indentation.

grails-app/migrations/changesets/changelog-11.0.xml
```diff
@@ -0,0 +1,41 @@
...
+        <addColumn tableName="consent_collection_link">
+            <column name="require_mta" type="VARCHAR(255)"/>
+        </addColumn>
+    </changeSet>
+
+    <changeSet author="vvicario" id="changelog-11.0-2">
+        <addColumn tableName="storage_document">
+            <column name="consent_key" type="VARCHAR(255)"/>
+        </addColumn>
+    </changeSet>
```
rushtong: Don't know how I missed this file in the review, but this goes along with previous commentary - we don't need this new column. Consent key is stored in `issue_key`. Remember that Consent Groups are just `Issue` objects.

grails-app/migrations/changesets/changelog-11.0.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <changeSet author="vvicario" id="changelog-11.0-3">
+        <dropNotNullConstraint columnName="project_key"
+                               columnDataType="VARCHAR(255)"
+                               tableName="storage_document"/>
+    </changeSet>
+</databaseChangeLog>
```
rushtong: One thing this changeset does **not** do, and probably should, is add a nullable column for the id of consent collection link. I may be misunderstanding things a bit - which could be a naming problem.

vvicario: Agree, we missunderstood, working on change it, I'll let you know when changes are done. Thanks

vvicario:
@rushtong Greg, if you have time, can you re-review this? Thanks!

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -45,51 +46,34 @@ class NewConsentGroupController extends AuthenticatedController {
...
-                        sampleCollectionIds.each {
-                            new ConsentCollectionLink(
-                                    projectKey: source.projectKey,
-                                    consentKey: consent.projectKey,
-                                    sampleCollectionId: it,
-                                    creationDate: new Date()
-                            ).save(flush: true)
-                        }
-                    }
+                    consentCollectionLink.save(flush: true)
```
rushtong: Nit: save operations should be in a service class ... sorry if I missed this on the first review.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -101,6 +85,7 @@ class NewConsentGroupController extends AuthenticatedController {
             if (consent != null) {
                 issueService.deleteIssue(consent.projectKey)
             }
+            consentCollectionLink?.delete(flush: true)
```
rushtong: nit: same feedback here, anything that hits the database should be in a service class method that can also be wrapped in a transaction.

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -163,21 +163,22 @@ class StorageProviderService implements Status {
      * @param files The multipart files
      * @return
      */
-    StorageDocument saveMultipartFile(String displayName, String userName, String issueKey, String type, MultipartFile file) {
+    StorageDocument saveMultipartFile(String displayName, String userName, String issueKey, String type, MultipartFile file, Long consentCollectionLinkId) {
         StorageDocument document = new StorageDocument(
-                projectKey: issueKey,
+                projectKey: consentCollectionLinkId == null ? issueKey : null,
```
rushtong: This doesn't look right. This adds a layer of business logic to the method that doesn't belong here. If this method is handed both a `projectKey` and a `consentCollectionLinkId`, then this method should honor what was asked of it and insert both values as they were provided. It might not make logical sense from a larger application perspective, but that business logic should be elsewhere, not here where we're storing the document.

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -238,8 +239,8 @@ class StorageProviderService implements Status {
         if (!document.mimeType) {
             throw new IllegalArgumentException("Mime Type is required")
         }
-        if (!document.projectKey) {
-            throw new IllegalArgumentException("Project Key is required")
+        if (!document.projectKey && !document.consentCollectionLinkId) {
+            throw new IllegalArgumentException("Project Key or consentCollectionLink must be specified")
```
rushtong: Why are we enforcing this here? What value does this provide to the application to actively prevent these two values from being set at the same time? Let me add here that there is no **business requirement** that this restriction be in place.

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -163,21 +163,22 @@ class StorageProviderService implements Status {
...
                 username: userName,
                 creationDate: new Date(),
-                status: DocumentStatus.PENDING.status
+                status: DocumentStatus.PENDING.status,
+                consentCollectionLinkId: consentCollectionLinkId
         )
         if (saveStorageDocument(document, file.getInputStream())) {
             persistenceService.saveEvent(
-                    document.projectKey,
+                    issueKey,
```
rushtong: It could be the case that this is null now that project and ccl_id are each nullable values.

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
vvicario: ok I can remove this validation, the idea was to validate that project key or consent key are not empty, if project key and consent collection link are null at the same time, doesn't make any sence to save a document without any reference.

rushtong: Pretty sure that I misread this code before - I thought you were checking to ensure that **only** one or the other was populated, but not both of them populated together. This is what I meant as not being a "business requirement". 

What you describe **does** make sense, there should be at least one OR the other populated, both is OK, but both empty is not OK. So, please accept my apologies ... we should have this error condition in place.

grails-app/services/org/broadinstitute/orsp/ConsentCollectionLinkService.groovy
```diff
@@ -0,0 +1,16 @@
...
+
+    ConsentCollectionLink save(ConsentCollectionLink consentCollectionLink) {
+        consentCollectionLink.save(flush: true)
+    }
+
+    void delete(ConsentCollectionLink consentCollectionLink) {
+        consentCollectionLink?.delete(flush: true)
+    }
+}
+
```
rushtong: We already have a class for this, [PersistenceService](https://github.com/broadinstitute/orsp-pub/blob/feature/grails-app/services/org/broadinstitute/orsp/PersistenceService.groovy)

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -208,12 +209,14 @@ class StorageProviderService implements Status {
...
-                    null
-            )
+            if (document.projectKey != null) {
+                persistenceService.saveEvent(
+                        document.projectKey,
+                        document.creator,
+                        "Adding document " + document.fileName + " to project",
+                        EventType.UPLOAD_DOCUMENT
+                )
+            }
```
rushtong: So  this is a good compromise in that we only save events for documents uploaded directly for an issue/consent group. We need some kind of TODO/ticket to follow up on saving an event for when a CCL document is uploaded too.

vvicario:
@rushtong  Can you please take a look at this and let me know if any changes needs to be made. Thanks

grails-app/services/org/broadinstitute/orsp/PersistenceService.groovy
```diff
@@ -45,4 +45,12 @@ class PersistenceService {
         ).save(flush: true)
     }
 
+    ConsentCollectionLink saveConsentCollectionLink(ConsentCollectionLink consentCollectionLink) {
+        consentCollectionLink.save(flush: true)
+    }
+
+    void deleteCollectionLink(ConsentCollectionLink consentCollectionLink) {
+        consentCollectionLink?.delete(flush: true)
+    }
```
rushtong: Thank you for these!

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
```diff
@@ -173,14 +173,15 @@ class StorageProviderService implements Status {
...
                 username: userName,
                 creationDate: new Date(),
-                status: DocumentStatus.PENDING.status
+                status: DocumentStatus.PENDING.status,
+                consentCollectionLinkId: consentCollectionLink?.id
         )
         if (saveStorageDocument(document, file.getInputStream())) {
             persistenceService.saveEvent(
-                    document.projectKey,
+                    document.projectKey == null ? consentCollectionLink.consentKey : document.projectKey,
```
rushtong: This poses an interesting question - and I agree with this solution - but the question is, what is the CCL document more closely associated with? I think this is the right answer.

grails-app/services/org/broadinstitute/orsp/StorageProviderService.groovy
rushtong: Thank you for adding this back in 😄

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -233,22 +272,41 @@ class NewConsentGroup extends Component {
...
-    let requireMta = false;
     let institutionalSourcesName = false;
     let institutionalSourcesCountry = false;
-
+    let noConsentFormReason = false;
     let isValid = true;
+    
+    if (!this.state.isConsentFormPresent && isEmpty(this.state.generalDataFormData.noConsentFormReason)) {
+      noConsentFormReason = true;
+      isValid = false;
```
rushtong: There is an interesting pattern here. Instead of assigning and potentially re-assigning `isValid` many, many times, instead, it could be calculated once at the end of this sequence of conditional checks. For example, 
```
const isValid = !(noConsentFormReason || consentGroupName || investigatorLastName || institutionProtocolNumber || institutionProtocolNumber || requireMta || collaboratingInstitution || describeConsentGroup || institutionalSourcesName || institutionalSourcesCountry || institutionalSourcesName);
```
This refactoring is a bit outside the scope of the code here, but it might be worth looking into.

vvicario:
I agree with you Greg, I'll change it in my current branch. Thank you!

PR #156, opened 2019-05-20
Btrx 670 dgil - Implement Read Only view role across views

dags:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-670

## Changes
Actions that change, updating or inserting, information are prevented either hiding or disabling action starters when the only role assigned to the user is "ro_admin". 

This PR covers only the UI portion of this story: we need to add restrictions on the backend to prevent a "ro_admin" user to make any modification.

## Testing
Create / Update a user with only 'ro_admin' role in supplemental_role table. User should be able to see all the information but should not be able to edit or add anything.

---


grails-app/taglib/org/broadinstitute/orsp/AuthTagLib.groovy
```diff
@@ -61,6 +61,25 @@ class AuthTagLib {
         }
     }
 
+    /**
+     * Renders the tag body when the user does not have ro_admin role
```
rushtong: nit: I think this is the opposite of what the function is doing. It looks like it is printing the body content when the user **does have the role in question**

grails-app/taglib/org/broadinstitute/orsp/AuthTagLib.groovy
```diff
@@ -61,6 +61,25 @@ class AuthTagLib {
...
+     */
+    def isViewer = { attrs, body ->
+        if (SupplementalRole.isViewer(session["roles"] as Collection<String>)) {
+            out << body()
+        }
+    }
+
+    /**
+     * Renders the tag body when the user have ro_admin role
```
rushtong: See comment above, I think these descriptions are flipped.

src/main/webapp/components/Documents.js
```diff
@@ -75,7 +75,6 @@ export const Documents = hh(class Documents extends Component {
       showRemoveDocuments: !this.state.showRemoveDocuments,
       documentToRemove: row
     });
-    console.log("veronica")
```
rushtong: Nice catch

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -746,6 +751,35 @@ class ConsentGroupReview extends Component {
     });
   };
 
+  shouldShowThis = (label) => {
```
rushtong: I am not a fan of putting this logic here. If it's **just** here in UI code, then the APIs will never know about it. Not sure how much work it would be, but the objects being returned via the API should also know if a thing is editable or not. 

See also the duplicate code in ProjectReview.js. If we do keep this purely UI, then it should be a utility function instead.

grails-app/taglib/org/broadinstitute/orsp/AuthTagLib.groovy
dags: yes, comments are the ooposite in isViewer and isNotViewer. Will fix it.

grails-app/taglib/org/broadinstitute/orsp/AuthTagLib.groovy
dags: same. Agreed.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
dags: Agree. Ill review backend behavior in all cases.

dags:
Since there are other roles that need some actions restricted in the backend, and present similar security risks today, we'll create a new story to implement roles restrictions properly in the backend.

PR #157, opened 2019-05-20
[BTRX-674] Add Existing CG and Sample Collection

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-674

## Changes
New view to create a link between a _Sample/Data Cohort_ and a _Sample Collection_ within the Project context. Answering _International Cohorts_, _Security_ and _MTA_ questions. Also is possible to upload documents.

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/SampleConsentLinkController.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+    def pages() {
+        render(view: "/linkWizard/index")
+    }
+
+    def save() {
+        JsonParser parser = new JsonParser()
+        User user = getUser()
+        List<MultipartFile> files = request.multiFileMap.collect { it.value }.flatten()
+        ConsentCollectionLink consentCollectionLink = IssueUtils.getJson(ConsentCollectionLink.class, parser.parse(request.parameterMap["dataConsentCollection"].toString())[0])
```
rushtong: This looks a little brittle by assuming there is a 0-th element in the parsed parameter map. Some error handling around this would be safer.

grails-app/controllers/org/broadinstitute/orsp/api/SampleConsentLinkController.groovy
```diff
@@ -0,0 +1,44 @@
...
+                    storageProviderService.saveMultipartFile(user.displayName, user.userName, null, it.name, it, consentCollectionLink)
+                }
+            }
+
+            response.status = 201
+            render([message: consentCollectionLink] as JSON)
+        } catch (Exception e) {
+            flash.error = e.getMessage()
+            response.status = 500
+            render([error: e.getMessage()] as JSON)
```
rushtong: If you hit an error partway through a list of files being uploaded, there will be `N` files in an invalid state and a partially saved `consentCollectionLink`. I think this could use some better cleanup handling.

grails-app/views/linkWizard/index.gsp
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
rushtong: Is there a reason we have to have this as a separate react app outside of where these questions lie now? I worry about us propagating this anti-pattern unless absolutely necessary. If possible, reference the `LinkWizard` component in the current project wizard. 

Caveat - I have not looked at this in the UI so I don't know if the above advice is possible or not. What I do know is that we are increasing the technical debt of [BTRX-609](https://broadinstitute.atlassian.net/browse/BTRX-609) every time we do this. It would be best to make an effort to integrate this now if possible.

src/main/webapp/linkWizard/LinkWizard.js
```diff
@@ -0,0 +1,416 @@
...
+      prev.linkFormData = updatedForm;
+      return prev;
+    }, () => {
+      this.isValid(field);
+    })
+  };
+
+  componentDidCatch(error, info) {
+    console.log('----------------------- new consent group error ----------------------');
+    console.log(error, info);
```
rushtong: Have we agreed on this pattern (and if so, I think it's a mistake to keep doing this)? I would prefer that we don't log errors this way and instead do something smarter, like inform the user that an error happened.

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -0,0 +1,200 @@
...
+  };
+
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true }
+  }
+
+  componentDidCatch(error, info) {
+    console.log('----------------------- error ----------------------');
+    console.log(error, info);
```
rushtong: See other comment about error logging

webpack.config.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
     project: './src/main/webapp/project/index.js',
     projectDocument: './src/main/webapp/projectDocument/index.js',
     projectReview: './src/main/webapp/projectReview/index.js',
-    search: './src/main/webapp/search/index.js'
+    search: './src/main/webapp/search/index.js',
+    linkWizard: './src/main/webapp/linkWizard/index.js'
```
rushtong: This should be alphabetized for clarity. We've had cases where duplicates were defined, but out of order, so it was harder to see.

grails-app/controllers/org/broadinstitute/orsp/api/SampleConsentLinkController.groovy
tadeork: I think it's enough if I just move this line to the try-catch block, right?

grails-app/controllers/org/broadinstitute/orsp/api/SampleConsentLinkController.groovy
tadeork: Sure, good idea.

grails-app/views/linkWizard/index.gsp
tadeork: I think this will change when we implement all the application in a react-way. As I understand this should live as a separate app meanwhile.

src/main/webapp/linkWizard/LinkWizard.js
tadeork: We have been using this pattern since we started working with React in ORSP. This has been replicated in every new component. I agree is not the best one, we just have not prioritized, defined or discussed a better, generic approach to handle errors. We should define how / how much info to include in an error page. Surly add a story to improve this.

src/main/webapp/linkWizard/SelectSampleConsent.js
rushtong: See https://broadinstitute.atlassian.net/browse/BTRX-684

grails-app/views/linkWizard/index.gsp
rushtong: To get a better handle on this, I spun up this branch and looked around. 

1. There is a link to this wizard from the list of consent groups in a project
2. The consent group tab is populated with an ajax call to the controller via jquery: `$("#consent-groups").load(...`
3. The controller renders a gsp which is not converted yet. 

The specific story this PR will impact is: [BTRX-613](https://broadinstitute.atlassian.net/browse/BTRX-613)
The net effect is that we're increasing the scope of that ticket (or any other ticket) when we add new functionality this way.

rushtong:
There is a noticeable delay when the content of `Select Sample/Data Cohort` populates in step 1 of the wizard. Not sure if there's a way to show a loading indicator in the drop down, but it was confusing that at first there was nothing to select, then a minute later, there were results to select. Note the 12 second delay in getting the consent groups - we need to file a technical debt story to improve that performance - it's terrible. 

Screen Shot 2019-05-21 at 8 05 16 AM

rushtong:
There is a bug in that it doesn't let me choose a sample collection that is already associated to a consent group. The rule for associating  consent groups + sample collections are:

1. It is possible to add a consent group + a sample collection that has not been associated to a consent group already. I think this is working in this branch.
2. It is possible to add a consent group + a sample collection that is associated to the selected consent group. This is what I cannot replicate in this branch.

`SC-11795 - Promega - K562 Control` is in `ORSP-590`: 
Screen Shot 2019-05-21 at 8 11 04 AM

`SC-11795 - Promega - K562 Control` is not a selectable option when adding that consent group to a project:
Screen Shot 2019-05-21 at 8 10 57 AM

src/main/webapp/linkWizard/LinkWizard.js
rushtong: Thank you!

grails-app/views/linkWizard/index.gsp
tadeork: I totally agree with you about this caveat.
Is there another way to implement this view in the meantime?

wloforteBroad:
@rushtong I created a ticket to add the Sample Collection association case we are not handeling in this PR. Please see: https://broadinstitute.atlassian.net/browse/BTRX-685. We will work on this ticket in this upcoming sprint.

grails-app/views/linkWizard/index.gsp
rushtong: I think what you have here @tadeork is the best interim solution. I was hoping that the wizard component could be called from another, pre-existing react component that was already written. I had to run it locally to realize that would not work.

PR #160, opened 2019-05-21
Btrx 682 sample collection page

LeoForconesi:
## Addresses
[BTRX-682](https://broadinstitute.atlassian.net/browse/BTRX-682) : ORSP: Associated Sample Collections Page

## Changes
- Added a new query to find all Sample Collection links corresponding to the pair Consent group / Project Key
- Modify review components to adapt the requested info
- New DTO object for Ui purposes. The new query mentioned above, will map to this new DTO type of object.

## Testing
Create a new project, add a new sample/data cohort selecting an existing sample collection. After their creation you can access the new page InfoLink from this new sample/data cohort's New Docuemnts tab, at the Associated project table at the bottom, through the link next to the project name listed in it.

---


grails-app/controllers/org/broadinstitute/orsp/api/InfoLinkController.groovy
```diff
@@ -0,0 +1,33 @@
...
+    def showInfoLink() {
+        render(view: "/infoLink/index")
+    }
+
+    def getProjectSampleCollections() {
+        String consentKey = params.consentKey
+        String projectKey = params.projectKey
+        try {
+            Gson gson = new Gson()
+            def result = queryService.findCollectionLinksByConsentKeyAndProjectKey(consentKey, projectKey)
```
rushtong: This should be a typed object, not a `def`. I think you want `Map<ConsentCollectionLinkDTO, List<StorageDocument>>` here.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -328,6 +329,31 @@ class QueryService implements Status {
...
+                        ' c.text_compliance textCompliance, c.require_mta requireMta, c.sample_collection_id sampleCollectionId, ' +
+                        ' sc.name collectionName, ic.summary consentName, ip.summary projectName, c.international_cohorts internationalCohorts ' +
+                        ' from consent_collection_link c ' +
+                        ' inner join issue ic on ic.project_key = c.consent_key ' +
+                        ' inner join issue ip on ip.project_key = c.project_key ' +
+                        ' left join sample_collection sc on sc.collection_id = c.sample_collection_id' +
+                        ' where c.project_key = :projectKey' +
+                        ' and c.consent_key = :consentKey'
+        List<ConsentCollectionLinkDTO> results = session.createSQLQuery(query)
+                .setResultTransformer(Transformers.aliasToBean(ConsentCollectionLinkDTO.class))
```
rushtong: I don't think I've ever used `Transformers. aliasToBean` before. Not sure how that compares to how we do it elsewhere in this class.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -328,6 +329,31 @@ class QueryService implements Status {
...
+                .setString('projectKey', projectKey)
+                .setString('consentKey', consentKey)
+                .list()
+        results.collect {
+            sampleInfo.put(it, StorageDocument.findAllByConsentCollectionLinkId(it.id))
```
rushtong: This is a nested query which will degrade performance really fast. Better to first pull the all of the ids into a single collection then query all storage documents with those ids. I also strongly recommend we do not use the GORM methods here (i.e. `StorageDocument.findAllByConsentCollectionLinkId`) ... they are notoriously slow. Much better to use raw sql to get all of the storage documents and then do something like `addEntity` over the results if you want to populate them.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -100,6 +100,17 @@ class ConsentGroupDocuments extends Component {
     return [this.props.serverURL, projectType, "show", projectKey,"?tab=review"].join("/");
   };
 
+  redirectToInfoLink = (projectKey) => {
+    let key = projectKey.split("-");
+    let projectType = '';
+    if (key.length === 3) {
+      projectType = key[1].toLowerCase();
```
rushtong: This is not always true. There are quite a few legacy projects that are `ORSP-X` (and many dev projects start with `DEV-`) and you won't get the right value here. Better to get the type from the full project object itself rather than trying to guess what it might be.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: There wasn't any existent representative object that we could use to be returned to the ui. We considered that in order to not add unnecessary or trash info to a domain object, we thought it would be better a DTO object.

On the other hand, since we weren't using a domain entity, we needed to map somehow what we got from the query to the new DTO object, so as result from a little research, we hit to this solution(`Transformers. aliasToBean`), that will map any given property passed as an alias. 

Does this make sense to you? Please let me know if it needs further explanation.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: I've created a new query that receives a list of ids to get all documents, then i build the map by matching consent collections links id and storage documents id.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
LeoForconesi: You're right! either way I removed this part of the code since this wasn't necessary here. I will take into consideration this comment though.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
rushtong: Sure ... works for me.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -349,11 +351,35 @@ class QueryService implements Status {
                 .setString('consentKey', consentKey)
                 .list()
         results.collect {
-            sampleInfo.put(it, StorageDocument.findAllByConsentCollectionLinkId(it.id))
+            consentCollectionIds.add(it.id)
+            consentCollectionLinkList.add(it)
+        }
+        List <StorageDocument> storageDocuments = findAllDocumentsBySampleCollectionId(consentCollectionIds)
+        consentCollectionLinkList.collect {
```
rushtong: I think you want an `each` here. `collect` returns a collection of items that you're not really using here.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -349,11 +351,35 @@ class QueryService implements Status {
...
+            sampleInfo.put(it,
+                    storageDocuments.findAll{
```
rushtong: This is also a nested query ... i.e., you're executing 1 query per item in `consentCollectionLinkList`. I think you just moved the problem to a new location in the code rather than fixing the problem.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -349,11 +351,35 @@ class QueryService implements Status {
...
+                    doc -> doc.consentCollectionLinkId == it.id
+                }
+            )
         }
         sampleInfo
     }
 
+    List <StorageDocument> findAllDocumentsBySampleCollectionId(List <Integer> consentCollectionIds) {
```
rushtong: What I think you really want here is not a list of storage documents. You want a map of consent collection link id -> list of storage documents. That will allow you to get rid of the nested query above. You can easily do this with the collection `groupBy` convenience method.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: Oh! eitherway these are not mysql queries if that's what you meant here. Made changes to remove findAll.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: Great idea. I had to cast a value to BigInteger here to match types from the place this function is called.

LeoForconesi:
@rushtong I've just reviewed your comments.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -328,6 +329,54 @@ class QueryService implements Status {
...
+                ' select * ' +
+                ' from storage_document ' +
+                ' where consent_collection_link_id in :consentCollectionIds '
+        final SQLQuery sqlQuery = session.createSQLQuery(query)
+        final results = sqlQuery.with {
+            addEntity(StorageDocument)
+            setParameterList('consentCollectionIds', consentCollectionIds)
+            list()
+        }
+        results.groupBy { it.consentCollectionLinkId as BigInteger }
```
rushtong: I'm a little confused by the cast ... the value is a `Long` on `StorageDocument` and should have been a `Long` value when coerced with `addEntity(StorageDocument)`. 

I think maybe there's something funky going on with `.setResultTransformer(Transformers.aliasToBean(ConsentCollectionLinkDTO.class))` because that's where I think you are expecting a `BigInteger`? 

In any case, this is a minor point - I'm OK with it as is. It would be nice if the typing could be cleaned up, but don't let that hold up this PR.

src/main/groovy/org/broadinstitute/orsp/ConsentCollectionLinkDTO.groovy
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.orsp
+import org.broadinstitute.orsp.IssueType
+
+class ConsentCollectionLinkDTO {
+
+    BigInteger id
```
rushtong: Ah ... here I think this should be a Long to match how it is defined in `StorageDocument`

src/main/webapp/components/SampleCollectionWizard.js
```diff
@@ -0,0 +1,133 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      currentStepIndex: 0,
+    };
+  }
+
+  componentDidCatch(error, info) {
+    console.log('----------------------- error ----------------------');
+    console.log(error, info);
```
rushtong: Note for @vvicario that her current error story will be impacted by this.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
rushtong: Looks like this is still an issue in this PR. Is there a commit missing?

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
vvicario: IMO: I think we can work on that on https://broadinstitute.atlassian.net/browse/BTRX-686, we have to refactor all this code, we will show only one collection per time, so results will contain only one result

src/main/groovy/org/broadinstitute/orsp/ConsentCollectionLinkDTO.groovy
vvicario: If we put Long, we will get an error, can not be casted, this is why we defined it as BigInteger

src/main/webapp/components/SampleCollectionWizard.js
vvicario: sure, I will remove it. I already removed them everywhere in my current branch

LeoForconesi:
@rushtong Could you please review again?

src/main/groovy/org/broadinstitute/orsp/ConsentCollectionLinkDTO.groovy
rushtong: I understand that you're seeing a cast error, but that error should be **because** this is not a Long. My point is that the cast is only fixing a symptom of the real problem, and I think this is the source of the real problem.

src/main/webapp/components/SampleCollectionWizard.js
rushtong: I'm not asking for this to be removed in this PR, just noting for the record that your error story will need to look at this instance too.

rushtong:
Will be out for the evening - will check again tomorrow.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -374,7 +373,7 @@ class QueryService implements Status {
             setParameterList('consentCollectionIds', consentCollectionIds)
             list()
         }
-        results.groupBy { it.consentCollectionLinkId as BigInteger }
+        results.groupBy { it.consentCollectionLinkId }
```
rushtong: Yes, exactly ... this is much cleaner! Using `Object`, `def`, and casting are obvious [code-smells](https://martinfowler.com/bliki/CodeSmell.html) that indicate there is probably a better way of doing what you want.

PR #161, opened 2019-05-22
[BTRX-688] Document Types in Add Sample

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-688

## Changes
Now the documents list type is the same as new Sample/Data Cohort. And minor style correction.

## Testing

---

### DIFF
--- a/src/main/webapp/linkWizard/SelectSampleConsent.js
+++ b/src/main/webapp/linkWizard/SelectSampleConsent.js
@@ -1,5 +1,5 @@
 import { Component } from 'react';
-import { hh, button, div } from 'react-hyperscript-helpers';
+import { hh, button, div, p } from 'react-hyperscript-helpers';
 import { WizardStep } from "../components/WizardStep";
 import { InputFieldSelect } from "../components/InputFieldSelect";
 import { Panel } from '../components/Panel';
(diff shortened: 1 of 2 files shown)


src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -109,7 +109,7 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
 
   loadOptions() {
     let documentOptions = [];
-    DOCUMENT_TYPE.forEach(type => {
+    CONSENT_DOCUMENTS.forEach(type => {
```
rushtong: This could be more efficient ... instead of mutating a variable, you can assign it to the result of the `map` operation, which already returns a collection of whatever is defined inside the map. Something like this: 
```
const documentOptions = CONSENT_DOCUMENTS.map(type => { value: type, label: type });
```
In general, try to [avoid mutation and instead use constants](https://medium.com/poka-techblog/simplify-your-javascript-use-map-reduce-and-filter-bd02c593cc2d).

tadeork: I agree, good suggestion. I'm going to do the same also in my actual story.

PR #165, opened 2019-05-27
[BTRX-691] Scroll top wizards

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-691

## Changes
The only change needed was to add line in the Wizard Step component.

## Testing

---

### DIFF
--- a/src/main/webapp/components/Wizard.js
+++ b/src/main/webapp/components/Wizard.js
@@ -38,9 +38,9 @@ export const Wizard = hh(class Wizard extends Component {
   };
 
   nextStep = (e) => {
-    window.scroll(0, 0);
     e.preventDefault();
     if (this.props.isValid(this.state.currentStepIndex, null)) {
+      window.scroll(0, 0);
       this.setState(prev => {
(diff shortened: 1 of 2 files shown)


src/main/webapp/components/Wizard.js
```diff
@@ -35,6 +35,7 @@ export const Wizard = hh(class Wizard extends Component {
   nextStep = (e) => {
     e.preventDefault();
     if (this.props.isValid(this.state.currentStepIndex, null)) {
+      window.scroll(0, 0);
```
rushtong: nit: this change would be more isolated and less likely to impact different browser configurations if it scrolled to the top of the component instead of going to the top of the window.

LeoForconesi: Hey @rushtong! , how about an Util function, browser compatible scrollToTop() to handle and re-use elsewhere? Does this change covers your request?

rushtong: That would definitely be an improvement, thank you!

PR #167, opened 2019-05-27
[BTRX-685] Select sample already associated

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-685

## Changes
The Sample Collection select component now show a heading for the already linked SC, and another heading for the non linked SC.

## Testing
Describe for the reviewer how to test the specifics of this PR, both positive and negative cases. Also both selects now displays a loading indicator.

---

### DIFF
--- a/grails-app/views/layouts/main_component.gsp
+++ b/grails-app/views/layouts/main_component.gsp
@@ -114,7 +114,7 @@
         attachDocumentsURL: "${createLink(uri: '/api/files-helper/attach-document', method: 'POST')}",
         consentGroupUrl: "${createLink(controller: 'newConsentGroup', action: 'findByUUID')}",
         consentNamesSearchURL: "${createLink(controller: 'consentGroup', action: 'consentGroupSummaries')}",
-        getConsentGroups: "${createLink(controller: 'consentGroup', action: 'getConsentGroups')}",
+        consentGroupsUrl: "${createLink(controller: 'consentGroup', action: 'getConsentGroups')}",
         createConsentGroupURL: "${createLink(controller:'newConsentGroup', action: 'save', uri: '/api/consent-group', method: 'POST')}",
         createProjectURL: "${createLink(controller:'project', action: 'save', uri: '/api/project', method: 'POST')}",
         error: "${error}",
@@ -134,8 +134,8 @@
         serverURL: "${webRequest.baseUrl}",
         userNameSearchUrl: "${createLink(controller: 'search', action: 'getMatchingUsers')}",
         sourceDiseases: "${createLink(controller: 'search', action: 'getMatchingDiseaseOntologies')}",
-        unConsentedSampleCollections: "${createLink(controller: 'consentGroup', action: 'unConsentedSampleCollections')}",
-        getConsentGroupSampleCollections: "${createLink(controller: 'consentGroup', action: 'getConsentGroupSampleCollections')}",
+        unlinkedSampleCollectionsUrl: "${createLink(controller: 'consentGroup', action: 'unConsentedSampleCollections')}",
+        linkedSampleCollectionsUrl: "${createLink(controller: 'consentGroup', action: 'getConsentGroupSampleCollections')}",
       };
     </script>
 
--- a/grails-app/views/layouts/main_component.gsp
+++ b/grails-app/views/layouts/main_component.gsp
@@ -133,6 +133,8 @@
         serverURL: "${webRequest.baseUrl}",
         userNameSearchUrl: "${createLink(controller: 'search', action: 'getMatchingUsers')}",
         sourceDiseases: "${createLink(controller: 'search', action: 'getMatchingDiseaseOntologies')}",
+        unConsentedSampleCollections: "${createLink(controller: 'consentGroup', action: 'unConsentedSampleCollections')}",
+        getConsentGroupSampleCollections: "${createLink(controller: 'consentGroup', action: 'getConsentGroupSampleCollections')}",
       };
     </script>
 
--- a/src/main/webapp/linkWizard/LinkWizard.js
+++ b/src/main/webapp/linkWizard/LinkWizard.js
@@ -305,6 +305,7 @@ export const LinkWizard = hh( class LinkWizard extends Component {
     // if (field === "consentGroup") {
     //   this.getAllSampleCollections(updatedForm.key);
     // }
+    console.log('updatedForm', updatedForm, 'field', field);
     if (this.state.currentStep === 0) {
       this.validateLinkStep(field);
     }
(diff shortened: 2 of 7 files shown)


grails-app/views/layouts/main_component.gsp
```diff
@@ -134,6 +134,8 @@
         serverURL: "${webRequest.baseUrl}",
         userNameSearchUrl: "${createLink(controller: 'search', action: 'getMatchingUsers')}",
         sourceDiseases: "${createLink(controller: 'search', action: 'getMatchingDiseaseOntologies')}",
+        unConsentedSampleCollections: "${createLink(controller: 'consentGroup', action: 'unConsentedSampleCollections')}",
+        getConsentGroupSampleCollections: "${createLink(controller: 'consentGroup', action: 'getConsentGroupSampleCollections')}",
```
rushtong: Nit: naming. For lines 136-139, we've broken with the naming convention of having URL be part of the variable name. `unConsentedSampleCollections` sounds like an actual list of unconsented sample collections rather than the url that provides the list of unconsented sample collections. And while `getConsentGroupSampleCollections` is more clear, it also breaks with the pattern of ending with `Url`, as does `sourceDiseases` which is not technically part of this PR. The naming consistency is important so code that uses it is easier to understand. For example, later on in this PR, there is a property assignment: 
```
unConsentedSampleCollections: this.props.unConsentedSampleCollections
``` 
that suspiciously looks like a list of things rather than an API call to get the list of things.

rushtong: I'm seeing this more now that I look ... see `getConsentGroups`, etc. We should try to be more consistent as this gets confusing very fast.

src/main/webapp/linkWizard/LinkWizard.js
```diff
@@ -381,12 +336,13 @@ export const LinkWizard = hh( class LinkWizard extends Component {
...
           currentStep: currentStep,
-          existingConsentGroups: this.state.existingConsentGroups,
           consentGroup: this.state.consentGroup,
           updateForm: this.updateGeneralForm,
           projectKeyLabel: this.props.projectKey,
           consentGroupIsLoading: this.state.consentGroupIsLoading,
-          sampleCollectionIsLoading: this.state.sampleCollectionIsLoading,
+          getConsentGroupSampleCollections: this.props.getConsentGroupSampleCollections,
+          getConsentGroups: this.props.getConsentGroups,
+          unConsentedSampleCollections: this.props.unConsentedSampleCollections,
```
rushtong: All three of these are global variables ... why do they need to be properties passed into the component? Same thing goes for the other urls that are passed in as props. These should be available to the component regardless of how it is instantiated. In principle, everything that's in `main_component.gsp` should be accessible to any component. If that's not the case, there should be some refactoring done to make that happen.

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -109,6 +114,71 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
     this.setState({ documentOptions: documentOptions });
   };
 
+  getConsentGroups = () => {
+    this.setState({ consentGroupIsLoading: true });
+    ConsentGroup.getConsentGroupNames(this.props.getConsentGroups).then(
```
rushtong: Here's an example where naming is confusing. We have a component method and a property with the same name, `getConsentGroups`. They should be more clearly named so it's super clear what's happening to/with what.

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -109,6 +114,71 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
...
+    );
+  };
+
+  getAllSampleCollections = (consentKey) => {
+    this.setState({ sampleCollectionIsLoading: true });
+
+    SampleCollections.getCollectionsCGLinked(this.props.getConsentGroupSampleCollections, consentKey).then(
+      resp => {
+        const sampleCollectionList = [];
+        sampleCollectionList.push({label: "Sample Collections Linked to " + consentKey, options: []});
```
rushtong: This is confusing - you define a `const` and then immediately mutate it. Should it be a `let` instead?

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -109,6 +114,71 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
...
+      resp => {
+        const sampleCollectionList = this.state.sampleCollectionList.splice(0);
+        sampleCollectionList.push({label: "Link New Sample Collections to Sample Data/Cohort: " + consentKey, options: []});
+        sampleCollectionList[1].options = resp.data.map(item => {
+          return {
+            key: item.id,
+            value: item.collectionId,
+            label: item.collectionId + ": " + item.name + " ( " + item.category + " )",
+          };
+        });
```
rushtong: This block of code is almost identical to the one above it. The select option builder code should be pulled out into a utility function that expects a label and a collection, builds a new list of options, and returns that for the `setState` call.

src/main/webapp/linkWizard/LinkWizard.js
tadeork: Yes, you're totally right. Also I just told the guys that we should change the way we were doing it. All that we need is present in `component.xxxx`

src/main/webapp/linkWizard/SelectSampleConsent.js
tadeork: I think not. It's a convention to use every variable as constant if possible. In this case I do it this way because of this:

> Constants are not a restriction on the value itself, but on the variable's assignment of that value. In other words, the value is not frozen or immutable because of const, just the assignment of it.

[Source](https://github.com/getify/You-Dont-Know-JS/blob/master/es6%20%26%20beyond/ch2.md)

src/main/webapp/linkWizard/SelectSampleConsent.js
tadeork: Yes, I agree.

src/main/webapp/linkWizard/SelectSampleConsent.js
rushtong: Holy hell, I didn't realize that. I absolutely loathe this fact. It's even worse than what you quoted ... 
> The a variable doesn't actually hold a constant array; rather, it holds a constant reference to the array. The array itself is freely mutable.

rushtong: Even still, the author there recommends that you do not change `const` values once declared ... and he uses the same word I did, `confusing`
> My advice: to avoid potentially confusing code, only use const for variables that you're intentionally and obviously signaling will not change. In other words, don't rely on const for code behavior, but instead use it as a tool for signaling intent, when intent can be signaled clearly.

tadeork: Got it. No problem.

PR #168, opened 2019-06-03
[BTRX-638] Update DUR form

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-638

## Changes
Added new fields in form and review page of DUR form

## Testing

---


grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
```diff
@@ -56,6 +56,22 @@ class DataUseRestrictionParser {
...
+
+        if (dataUseRestriction.noRestriction) {
+            dataUseRestriction.generalUse = true
+        }
+
+        dataUseRestriction.collaborationInvestigators = IssueUtils.getBooleanForParam(params.collaborationInvestigators)
+        dataUseRestriction.publicationResults = IssueUtils.getBooleanForParam(params.publicationResults)
+        dataUseRestriction.genomicResults = IssueUtils.getBooleanForParam(params.genomicResults)
+        if (dataUseRestriction.genomicResults) {
+            dataUseRestriction.genomicSummaryResults = (String)params.genomicSummaryResults
```
rushtong: nit: casting. `params` is a [GrailsParameterMap](http://docs.grails.org/3.1.3/api/grails/web/servlet/mvc/GrailsParameterMap.html) which inherits from [AbstractTypeConvertingMap](http://docs.grails.org/3.1.3/api/grails/util/AbstractTypeConvertingMap.html) and therefore has some built in casting methods that should help clean up this kind of thing.

grails-app/views/dataUse/edit.gsp
```diff
@@ -259,7 +288,8 @@
                               name="controlSetOption"
                               id="controlSetOptionUnspecified"
                               <g:if test="${restriction.controlSetOption?.equals("Unspecified") && !create}">checked</g:if>
-                              value="Unspecified">Unspecified</label>
+                              value="Unspecified">Unspecified                <tr>
```
rushtong: nit: alignment. `<tr>` should, I think, be on the next line to align with the closing `</tr>`

grails-app/views/dataUse/edit.gsp
```diff
@@ -307,7 +337,7 @@
 
             <span>
                 <label for="populationRestrictions">Future use is limited to research involving a specific population
-                [<abbr title="Use of the data is limited to tudies of particular population.">RS-POP</abbr>]</label>
+                [<abbr title="Use of the data is limited to studies of particular population.">RS-POP</abbr>]</label>
```
rushtong: Nice catch!

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
rushtong: Aa a side note, I have come to distrust a lot of the dynamic accessors of map keys like this (i.e. `params.genomicResults`) and prefer to use the getters to get at the values and check for null if necessary. Using the null-safe operator (`?`) is helpful for sitations like this too.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseDTO.groovy
```diff
@@ -14,7 +14,7 @@ class DataUseDTO {
     String controlSetOption
     String gender
     Boolean pediatric
-    List<String> populationRestrictions
+    String populationRestrictions
```
rushtong: Is this really true? Have we restricted the autocomplete to be a single choice instead of multiple? It's hard to tell from the PR, but I don't see that specific change.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
tadeork: I quite don't get this. `IssueUtils` returns `true` or `false` if get "Yes" - "No". If I use, for instance, `params.get("generalUse")` I'll get "Yes" or "No", but it's needed a boolean value.

grails-app/views/dataUse/edit.gsp
tadeork: Actually there's no need for this tags.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
rushtong: Let me test some code ... I'll give you a better example. I know you need a string, but the cast is the code-smell here. I'm pretty sure there's a cleaner way to do this without a hard cast.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseDTO.groovy
tadeork: It's true. This come from a change requested in the lucidchart. From a list to a free text. And then it was asked to get back to a list of ontologies.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
rushtong: Ahhh ... this is the real problem:
```
static DataUseRestriction fromParams(DataUseRestriction dataUseRestriction, Object params)
```
I thought this was a grails parameter map, but it's not, it's a plain `Object`. In DataUseController.save() ... you have a `GrailsParameterMap` that is passed in as the object. In `fromParams`, you change that from `GrailsParameterMap` to `Object`. Because you do that, you're losing a ton of built in functionality that Grails does for you for free. You end up re-implementing functions that have already been implemented for you when you up-cast the map to an object.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseDTO.groovy
rushtong: Then I'm confused ... if it is as you say
> And then it was asked to get back to a list of ontologies.
Then why is this changing from a list to a single string?

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
rushtong: So, looking more closely at this class - there are two ways that this is used. One is that it gets a `GrailsParameterMap` and the other is that it gets a JSON object. I do think the two should be distinctly different, but that's a good amount of re-work that is outside of the scope of this PR. I suggest a tech debt story to clean up usages of this class.

src/main/groovy/org/broadinstitute/orsp/consent/DataUseDTO.groovy
tadeork: It's a mistake that I made because I lost track of the back and forth changes. Should stay as a list.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
tadeork: OK, agree. I'll create the story as a tech debt then.

tadeork: Here https://broadinstitute.atlassian.net/browse/BTRX-697

PR #177, opened 2019-06-10
Btrx 705 improve sample collection

jaime-bernales-broad:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-705

## Changes
- Link names and urls are fully displayed when you mouseover in a link (tooltip/title)
- Column width size is improved depending on the column content, assigning more width to columns with larger strings

## Testing
- Sample Collection Table
- Project > Documents Tab 
- Consent Group

---

### DIFF
--- a/src/main/webapp/components/Table.js
+++ b/src/main/webapp/components/Table.js
@@ -8,6 +8,18 @@ import { Btn } from './Btn';
 import './Table.css';
 import { handleRedirectToProject } from "../util/Utils";
 
+const styles = { 
+  statusWidth: '140',
+  fileTypeWidth: '170',
+  docVersionWidth: '90',
+  creatorWidth: '130',
+  infoLinkWidth: '96',
+  creationDateWidth: '140',
+  removeWidth: '45',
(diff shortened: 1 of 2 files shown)


src/main/webapp/components/Table.js
```diff
@@ -149,7 +158,29 @@ export const Table = hh(class Table extends Component {
               return <TableHeaderColumn key={header.name}
                 dataField={header.value}
                 dataFormat={this.formatStatusColumn}
-                dataSort={true}>{header.name}</TableHeaderColumn>
+                dataSort={true}
+                width={'140px'}>{header.name}</TableHeaderColumn>
```
rushtong: These hard-coded values should be in a configurable style object elsewhere. Something like
```
const styles = { 
  statusWidth: 140,
  fileTypeWidth: 170,
  etc.
}
```
and then you can refer to it directly, `styles.statusWidth`, etc.

rushtong: Also, pretty sure you don't need the px in the string - I think just using the raw number will default to the pixels.

PR #178, opened 2019-06-13
BTRX - 606 - Role Management Screen

LeoForconesi:
## Addresses
[BTRX-606](https://broadinstitute.atlassian.net/browse/BTRX-606): ORSP- Role Management Screen

## Changes
- Create Role Management page, and Role Management Edit modal
- Add neccesary MySql queries to :
  - Get all users with their respective roles 
  - Insert new roles to a given userId
  - Delete roles for a given userId
- Convert UserController to be a REST resource.
- Added new functionalities to UserService:
  - To verify if a given role list to assign contains valid roles
  - There are some roles that shouldn't be assigned together, so a new method to check this rule is added
- Unify "admin" and "orps" roles for the Ui. There were some places where it was asked if the user was admin and others places where orsp was asked instead. The solution was a unique Utility function that does this work.

## Testing
- Role management page is accessible for "orsp" users in the top nav "admin" option. Role edition is accesible from the pencil icon located in the user's row.
- Read Only role, cannot coexists with Admin or Compliance Office roles, and viceversa.
---


grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -41,6 +41,8 @@ class UrlMappings {
         '/api/dul-email-notification'(controller: 'dulNotify', action: 'sendNotifications', method: 'POST')
         '/api/clarification-request/'(controller: 'clarification', action: 'addClarificationRequest', method: 'POST')
         '/api/sample-consent-link'(controller: 'sampleConsentLink', action: 'save', method: 'POST')
+        '/api/get-users'(controller: 'user', action: 'getOrspUsers', method: 'GET')
+        '/api/edit-UserRole'(controller: 'user', action: 'editOrspUserRole', method: 'PUT')
```
dags: please use only lowercase in url

grails-app/services/org/broadinstitute/orsp/UserService.groovy
```diff
@@ -151,4 +152,48 @@ class UserService {
...
+        if (userId != null) {
+            User user = User.findById(userId)
+            if (user != null && validateRoles(rolesToAssign)) {
+                queryService.deleteOrspUserRoles(userId)
+                if (!rolesToAssign.isEmpty()) {
+                    queryService.updateOrspUserRoles(user, rolesToAssign)
+                }
+            } else {
+                log.error("Error while trying to modify roles to userId: ${userId}.")
+                throw IllegalArgumentException()
```
rushtong: It would be nice to know what the error is here. Can you pass that into the exception so it can bubble up? Also, it would be nice to know *why* the failure happened. Was the user null? What failed in the validation? etc.

grails-app/views/rolesManagement/index.gsp
```diff
@@ -0,0 +1,14 @@
...
+    </head>
+    <body>
+        <g:javascript>
+            const rolesConstants = {
+            };
+        </g:javascript>
+        <div id="rolesManagement"></div>
+        <asset:javascript src="build/rolesManagement.js"/>
+    </body>
+</html>
```
rushtong: Can we add a story (or update one that might exist) to cover converting this to react? This PR adds to the technical debt that we're trying to pay down in other stories.

src/main/webapp/adminOnly/AdminOnly.js
```diff
@@ -81,13 +81,9 @@ export const AdminOnly = hh(class AdminOnly extends Component {
...
 
   isCurrentUserAdmin() {
-    User.getUserSession(this.props.userSessionUrl).then(
-      resp => {
-        this.setState({ isORSP: resp.data.isORSP });
-      }
-    ).catch(error => {
-      this.setState(() => { throw error; });
-    });
+    isCurrentUserAdmin().then(response => {
```
rushtong: This naming is confusing ... the method name here is the same as the method it is in ... so it looks like a recursive call, but maybe it isn't?

src/main/webapp/components/AlertMessage.css
```diff
@@ -31,3 +31,20 @@
...
+}
+
+.alertMessage.alert-danger .error-icon .btnPrimary {
+  color:red;
+  background-color: transparent;
+}
+
+.alertMessage.alert-danger .error-icon .btnPrimary .glyphicon {
+  font-size: 14px;
+}
```
rushtong: nit: add newline

src/main/webapp/components/RoleManagementEdit.js
```diff
@@ -0,0 +1,151 @@
+import { Component, Fragment } from 'react';
+import { hh, h, button, div } from 'react-hyperscript-helpers';
+import { Modal, ModalHeader, ModalTitle, ModalFooter, ModalBody } from 'react-bootstrap';
+import { InputFieldCheckbox } from "./InputFieldCheckbox";
+import { ROLE, USER_ROLES } from '../util/roles';
```
rushtong: I'm digging through the code here to figure out what you're doing, and this is pretty confusing to me. You have two things (`USER_ROLES`, `ROLES`) that are really the same thing, roles that a user can have. Both are ordered the same, so I'm completely missing the point of the having the `ROLE` object at all. Probably outside the scope of this PR, but I think these two things should be merged into one data structure.

src/main/webapp/components/RoleManagementEdit.js
```diff
@@ -0,0 +1,151 @@
...
+        ]),
+        h(ModalFooter, {className: 'roles-cta'}, [
+          button({ className: "btn buttonTertiary", disabled: this.state.disableClearButton, onClick: this.clearSelection}, ["Clear"]),
+          button({ className: "btn buttonSecondary", onClick: this.props.closeModal()}, ["Cancel"]),
+          button({ className: "btn buttonPrimary", disabled: this.state.disableSubmitButton || this.state.showError, onClick: this.submit }, ["Submit"])
+        ])
+      ])
+    );
+  }
+});
```
rushtong: nit newline

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -349,11 +349,9 @@ class ConsentGroupReview extends Component {
   };
 
   isCurrentUserAdmin() {
-    User.isCurrentUserAdmin(this.props.isAdminUrl).then(
-      resp => {
-        this.setState({ isAdmin: resp.data.isAdmin });
-      }
-    );
+    isCurrentUserAdmin().then(response => {
```
rushtong: Ok, I see where this is coming from now `import { isCurrentUserAdmin, isEmpty } from "../util/Utils";` ... and the fact the parent method name is the same as the imported function is very confusing. Please rename one or the other so it doesn't look like a recursive call.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -239,9 +241,11 @@ export const ProjectReview = hh(class ProjectReview extends Component {
       });
   }
 
-  isAdmin = () => {
-    return this.props.isAdmin === "true";
-  };
+  isCurrentUserAdmin() {
+    isCurrentUserAdmin().then(response => {
```
rushtong: Same feedback here.

src/main/webapp/util/Utils.js
```diff
@@ -45,3 +48,13 @@ export const scrollToTop = () => {
 export const handleRedirectToProject = (serverURL, projectKey) => {
   return [serverURL, "project", "main?projectKey=" + projectKey + "&tab=review"].join("/");
 }
+
+export const obtainStringRole = (role) => {
+  return USER_ROLES.map(it => it.value === role);
+}
+
+export const isCurrentUserAdmin = () =>  {
```
rushtong: As I see usages of this, why don't we change the API call that gets the user to calculate this on the server side and add it to the user object? We call this method so many times, it seems very unnecessary. Every page should have the user in context, so if this value were already on the user object, we would not have to make an additional query to get it.

rushtong: Answering my own question here. We already do this. See `AuthenticatedController.groovy`:
```
    def getSessionUser() {
        User user = getUser()
        render([
            'displayName': user.displayName,
            'emailAddress': user.emailAddress,
            'userName': user.userName,
            'isAdmin': isAdmin(),
            'isORSP': isORSP(),
            "isViewer": isViewer(),
            'isComplianceOffice': isComplianceOffice(),
        ] as JSON)
    }
```
The user object we get back from the server already knows if the user is an admin already. I think this method, and all callers of it, should be removed and look at the user directly.

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
LeoForconesi: Done :+1:

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1186,4 +1186,34 @@ class QueryService implements Status {
...
+            list()
+        }
+        results
+    }
+
+     void updateOrspUserRoles (User user, ArrayList<String> newRoles) {
+        final session = sessionFactory.currentSession
+        final String query = ' insert into supplemental_role (version, role, user, user_id) values '
+        StringBuffer stringBuffer = new StringBuffer().append(query)
+        String queryAndValues = stringBuffer.append(newRoles.collect{ it -> return "(0, '${it}', '${user.userName}', ${user.id})"}.join(" , ")).toString()
```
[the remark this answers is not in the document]
LeoForconesi: Ouch. Thanks for catching this, now it's using prepared statement for setting these.

grails-app/services/org/broadinstitute/orsp/UserService.groovy
LeoForconesi: Could you please check if the [modification](https://github.com/broadinstitute/orsp-pub/pull/178/files#diff-0329d5293b4b710e0a17b621f105f08bR207) covers this request? I think it does. Otherwise please let me know.

src/main/webapp/adminOnly/AdminOnly.js
LeoForconesi: it is, now that I look at it carefully. Eitherway I removed all these methods everywhere in realtion to another [comment](https://github.com/broadinstitute/orsp-pub/pull/178#discussion_r293786287) below this PR

src/main/webapp/components/RoleManagementEdit.js
LeoForconesi: I've removed `ROLE` constant object, and changed `USER_ROLES` in order to use only one data structure. You can check it [here](https://github.com/broadinstitute/orsp-pub/pull/178/files#diff-fda238a7164082c3e3529305f1f0ab37R1).

src/main/webapp/util/Utils.js
LeoForconesi: The main idea of ​​creating a helper in the ui was to get these session permission to unify and similar roles `orsp` and` admin`, but I see that we can save of doing this by using `isAdmin ()` from groovy's `SupplementalRole` class . 
I had to do a little [change](https://github.com/broadinstitute/orsp-pub/pull/178/files#diff-56bd5adea21ec7add82600bafad8443eR12) there due to, as mentioned above, Orsp and Admin are the same, at least for now, so if you call `isAdmin()` it verifies if the user has admin or orsp roles.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1186,4 +1186,34 @@ class QueryService implements Status {
         documents
     }
 
+    Collection<User> getUsers() {
+        final session = sessionFactory.currentSession
+        final String query =
+                ' select * from user ' +
+                        ' order by id '
```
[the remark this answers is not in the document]
LeoForconesi: I've replaced this query following as example as it is done in [Fundings](https://github.com/broadinstitute/orsp-pub/blob/feature/grails-app/services/org/broadinstitute/orsp/QueryService.groovy#L146) report. There's a good difference! 
[Change here](https://github.com/broadinstitute/orsp-pub/pull/178/files#diff-5f5fe31c391c7f2fe48a01cbd5d75cc8R1239)

grails-app/views/rolesManagement/index.gsp
LeoForconesi: Here's a story for this request. 
[BTRX-712](https://broadinstitute.atlassian.net/browse/BTRX-712): ORSP - Convert Role Management

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
```diff
@@ -29,4 +35,52 @@ class UserController {
...
+            def output = [:]
+            output['id'] = it.id
+            output['userName'] = it.userName
+            output['displayName'] = it.displayName
+            output['emailAddress'] = it.emailAddress
+            output['createdDate'] = it.createdDate
+            output['updatedDate'] = it.updatedDate
+            output['lastLoginDate'] = it.lastLoginDate
+            output['roles'] = it.roles.stream().map{role ->
+                role.getRole()}.collect()
```
rushtong: Can be simplified ... `it.roles.collect{r -> r.getRole()}` Groovy introduced just about everything that javas streams do long before java got around to it.

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
```diff
@@ -29,4 +35,52 @@ class UserController {
...
+        }
+        catch(IllegalArgumentException e) {
+            response.status = 400
+            render([error: e.message] as JSON)
+        }
+        catch(Exception e) {
+            log.error("Error while trying to modify roles to userId: ${userId}." + e.message)
+            response.status = 500
+            render([error: e.message] as JSON)
+        }
```
rushtong: Might be safer catching `Throwable` here.

grails-app/domain/org/broadinstitute/orsp/SupplementalRole.groovy
```diff
@@ -9,7 +9,7 @@ class SupplementalRole {
 
     public static final List<String> CCO_USERS = ["sdonnell", "saltzman"]
     public static final List<String> ORSP_ROLES = [COMPLIANCE_OFFICE, ORSP]
-    public static final List<String> ADMIN_ROLES = [ADMIN]
+    public static final List<String> ADMIN_ROLES = [ADMIN, ORSP]
```
rushtong: 👍

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1209,4 +1209,78 @@ class QueryService implements Status {
...
+        }.join(" , ")).toString()
+
+        final SQLQuery sqlQuery = session.createSQLQuery(queryAndValues)
+        newRoles.eachWithIndex { String role, int i ->
+            sqlQuery.setString("role${i}", role)
+            sqlQuery.setString("userName", user.userName)
+            sqlQuery.setLong("userId", user.id)
+        }
+        sqlQuery.executeUpdate()
+    }
```
rushtong: This implementation seems very complex to me. It looks like it should be drop dead simple ... 
```
final String query = ' insert into supplemental_role (version, role, user, user_id) values (:version, :role, :user, :userId)'
newRoles.each { it -> 
  sqlQuery.setInteger("version", 0)
  sqlQuery.setString("role", it)
  sqlQuery.setString("userName", user.userName)
  sqlQuery.setString("userId", user.id)
  sqlQuery.executeUpdate()
} 
```

grails-app/services/org/broadinstitute/orsp/UserService.groovy
rushtong: 👍 It does, thank you!

grails-app/services/org/broadinstitute/orsp/UserService.groovy
```diff
@@ -151,4 +152,64 @@ class UserService {
...
+    }
+
+    /*
+     It would be nice to know what the error is here.
+      Can you pass that into the exception so it can bubble up?
+      Also, it would be nice to know why the failure happened.
+       Was the user null? What failed in the validation? etc.
+
+
+    */
```
rushtong: do we need this comment?

grails-app/views/consentGroupReview/_index.gsp
```diff
@@ -0,0 +1,22 @@
...
+  </head>
+  <body>
+    <g:javascript>
+      const consentGroupReviewComponent = {
+        consentKey:'${issue.projectKey}',
+        consentGroupUrl: "${createLink(controller: 'consentGroup', action: 'getConsentGroup')}",
+        rejectConsentUrl: "${createLink(controller: 'newConsentGroup', action: 'delete')}",
+        projectKey: "${params.projectKey}",
+        updateConsentUrl: "${createLink(controller: 'newConsentGroup', action: 'update')}",
+        approveConsentGroupUrl: "${createLink(controller: 'newConsentGroup', action: 'approveConsentGroup')}"
```
rushtong: Pretty sure that these are already in `main_component` ... do you need to redefine them here too?
* consentKey
* createConsentGroupURL
* rejectConsentUrl
* projectKey
* updateConsentUrl
* approveConsentGroupUrl

grails-app/views/rolesManagement/index.gsp
rushtong: Thank you!

src/main/webapp/components/Table.css
```diff
@@ -32,3 +32,39 @@
...
+}
+
+.pagination.custom-component {
+    margin-top: 20px;
+    float: right;
+}
+
+.custom-component .page-item .page-link {
+    cursor: pointer;
+}
```
rushtong: nit: newline

src/main/webapp/components/Table.js
```diff
@@ -145,6 +146,21 @@ export const Table = hh(class Table extends Component {
...
+            className: "btn btn-default btn-sm edit",
+            onClick: this.props.editRole(row),
+            disabled: !this.props.isAdmin,
+            title: "Edit"
+          }, [
+            span({className: "glyphicon glyphicon-pencil"}, [])
+          ]
+      )]);
+  };
+
```
rushtong: So, technically, this should really not be part of the table code. Table should be very generic - by adding lots of other-feature-specific-functionality to it, we're breaking encapsulation. What this should be is a function defined in the RoleManagement component that is passed down into Table so that table knows what to do with it, i.e., letting table say, use function X for `dataFormat` in the case of column `X`. I'm afraid we might be a little too far down this rabbit hole right now ... table is super content dependant and getting worse.

rushtong: One thought, perhaps for a tech-debt story, is to refactor Table so it doesn't have to deal with all of the headers on its own. Maybe if we passed in a list of header values (each one either as a value or a function), then table could iterate through them and set the props for each one.

src/main/webapp/rolesManagement/RolesManagement.js
```diff
@@ -0,0 +1,182 @@
...
+import { Spinner } from '../components/Spinner';
+import { TablePaginator } from "../components/TablePaginator";
+
+const tableHeaders =
+  [
+    { name: 'User Name', value: 'userName' },
+    { name: 'Display Name', value: 'displayName' },
+    { name: 'Email Address', value: 'emailAddress' },
+    { name: 'Roles', value: 'roles' },
+  ];
```
rushtong: This is the right idea ... each one of these should also have a function for `dataFormat` so that Table doesn't have to know about it. It would take some refactoring around how we're using Table in other places too.

src/main/webapp/rolesManagement/RolesManagement.js
```diff
@@ -0,0 +1,182 @@
...
+
+   render() {
+    return(
+      div({ className: "roles-management" },[
+        span({ style: styles.pageTitle}, ["Roles Management"]),
+        Table({
+          headers: tableHeaders,
+          isAdmin: this.state.isAdmin,
+          data: this.state.users,
+          serverURL: component.serverURL,
```
rushtong: This should probably not be a prop on Table and instead just be a constant that Table always knows about.

src/main/webapp/rolesManagement/RolesManagement.js
```diff
@@ -0,0 +1,182 @@
...
+          onSearchChange: this.onSearchChange,
+          onSortChange: this.onSortChange
+        }),
+        TablePaginator({
+          onPageChange: this.onPageChange,
+          currentPage: this.state.currentPage,
+          lastPage: this.state.lastPage
+        }),
+        RoleManagementEdit({
+          serverURL: component.serverURL,
```
rushtong: Ditto ... `RoleManagementEdit` should know about the serverURL regardless of where it is instantiated, so probably better off not as a prop.

src/main/webapp/rolesManagement/index.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import '../index.css';
+import ErrorHandler from '../components/ErrorHandler';
+import RolesManagement from "./RolesManagement";
+
+ReactDOM.render(
+  <ErrorHandler>
+    <RolesManagement
+      loadingImage = {component.loadingImage}
```
rushtong: This should be a global constant as well, it shouldn't be a prop.

src/main/webapp/util/ajax.js
```diff
@@ -199,16 +199,25 @@ export const DocumentHandler = {
...
 
-  isCurrentUserAdmin(url) {
-    return axios.get(url);
+  getUserSession(url) {
+    return axios.get(url)
   },
 
-  addExtraProperties(url, projectKey, data) {
-    return axios.post(url+ '?id=' + projectKey, data );
+  getAllUsers(serverUrl, query) {
```
rushtong: If this utility could get access to the global `serverUrl`, that would cut down on a ton of boilerplate that we have now.

webpack.config.js
```diff
@@ -14,6 +14,9 @@ module.exports = {
     linkWizard: './src/main/webapp/linkWizard/index.js',
     mainIndex: './src/main/webapp/main/main_index.js',
     project: './src/main/webapp/project/index.js',
+    projectDocument: './src/main/webapp/projectDocument/ProjectDocument.js',
+    projectReview: './src/main/webapp/projectReview/index.js',
+    rolesManagement: './src/main/webapp/rolesManagement/index.js',
```
rushtong: Thank you for keeping these alphabetic!

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
LeoForconesi: Thanks for the suggestion! it's now simplified.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: Cool, My idea was to not execute n times a database query. But now that i think of it, newRoles never is going to be larger than 2. This is now changed

grails-app/services/org/broadinstitute/orsp/UserService.groovy
LeoForconesi: Omg, sorry for that, i've must have copied that there for following up a suggestion. This is gone!

grails-app/views/consentGroupReview/_index.gsp
LeoForconesi: I wasn't paying attention to these, you're right! I've removed those from consentGroupReview Index, by previously check if they weren't in use, or replacing them with global constants.

src/main/webapp/components/Table.js
LeoForconesi: That makes sense. This component started to grow and gain functionality uncontrolled. Also if its ok, it would be really cool to use React Bootstrap Table 2.

src/main/webapp/util/ajax.js
LeoForconesi: I wasn't aware that ajax file had direct access to global constants. But after talking about it with the team, i realize that it is. I would leave this a minor tech debt since changing these (even if its simple) it would require a re-test of the entire application to save any possible bug.

src/main/webapp/components/Table.js
rushtong: No objections from me on updating the react bootstrap table. We're on `"react-bootstrap-table": "4.3.1"` now, not sure what it would take to migrate to `react-bootstrap-table2`. I would suggest that be a very separate task, hopefully for obvious reasons.

src/main/webapp/util/ajax.js
rushtong: I wasn't sure either ... thanks for looking into that!

src/main/webapp/components/Table.js
LeoForconesi: I've created a story for this task [BTRX-716](https://broadinstitute.atlassian.net/browse/BTRX-716)

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
LeoForconesi: I went through the code to see if I was missing any other possible error, so I included SQLException to UserService method.
I sticked to the idea of catching `Exception` at the end of the code for any un-logged problem but this shouldn't occur in this method.

LeoForconesi:
@rushtong this Pr is ready to be reviewed again. And I've created stories for all larger tasks out of this Pr's scope

rushtong:
Re-reviewing!

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
rushtong: RE: Sql exceptions ... the content of those exceptions should never be returned raw, i.e. `render([error: e.message] as JSON)`, because they can contain relatively sensitive information about the query, etc. What I generally do is log that info and return something generic to the user.

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
```diff
@@ -29,4 +37,53 @@ class UserController {
...
+            userService.editUserRoles(userId, rolesToAssign)
+            response.status = 200
+            render([message: 'Role Updated'] as JSON)
+        } catch(IllegalArgumentException e) {
+            response.status = 400
+            render([error: e.message] as JSON)
+        } catch(SQLException e) {
+            response.status = 500
+            log.error("Error trying to execute delete/insert for new roles: ${rolesToAssign} to userId: ${userId}. " + e.message)
+            render([error: e.message] as JSON)
```
rushtong: I suggest we don't return the full error message here. Better to log it and then send the user something generic so that raw sql doesn't go back to the user.

grails-app/services/org/broadinstitute/orsp/UserService.groovy
```diff
@@ -151,4 +154,57 @@ class UserService {
...
+            User user = User.findById(userId)
+            def validatedRoles = validateRoles(rolesToAssign)
+            if (user != null && !validatedRoles?.hasErrors) {
+                queryService.deleteOrspUserRoles(userId)
+                if (!rolesToAssign.isEmpty()) {
+                    queryService.updateOrspUserRoles(user, rolesToAssign)
+                }
+            } else {
+                log.error("Error while trying to modify roles to userId: ${userId}. ${validatedRoles.errorMessage}")
+                throw new IllegalArgumentException()
```
rushtong: nit: suggestion for what to add to this exception: `"Error while trying to modify roles for user ${user.displayName}"`

src/main/webapp/components/Table.js
```diff
@@ -156,12 +172,16 @@ export const Table = hh(class Table extends Component {
...
         search={this.props.reviewFlow}
         options={{
           paginationSize: this.props.paginationSize,
           paginationPosition: 'bottom',
-          sizePerPage: this.props.sizePerPage
+          sizePerPage: this.props.sizePerPage,
+          onSizePerPageList: this.props.onSizePerPageListHandler,
+          onPageChange: this.props.onPageChange,
+          onSearchChange: this.props.onSearchChange,
+          onSortChange: this.props.onSortChange
```
rushtong: Curious ... what happens to other implementations of `Table` that do not provide these props? Does the search page or other components need to do anything specific here?

src/main/webapp/components/TablePaginator.js
```diff
@@ -0,0 +1,102 @@
...
+    this.setState( prev => {
+      prev.pageCounter = position + this.state.pageCounter - 1;
+      return prev;
+    });
+  };
+
+  render() {
+    return (
+      ul({className: "pagination custom-component"}, [
+        li({ className: "page-item", title: "first page", onClick: (e) => { this.goFirstPage() }}, [
```
rushtong: nit: do you need the `e` if you're not using it?

rushtong: Also a nit: you're using different syntax across these page-item elements, i.e. `onClick: (e) => { this.goFirstPage() }}` vs `onClick: this.goPrevPage`

src/main/webapp/projectReview/index.js
```diff
@@ -0,0 +1,25 @@
...
+ReactDOM.render(
+  <ErrorHandler>
+    <ProjectReview
+      issue = {projectReviewValues.issue}
+      searchUsersURL = {projectReviewValues.searchUsersURL}
+      projectKey = {projectReviewValues.projectKey}
+      projectUrl = {projectReviewValues.projectUrl}
+      addExtraPropUrl = {urls.saveExtraPropUrl}
+      isViewer = {projectReviewValues.isViewer}
+      serverURL = {projectReviewValues.serverURL}
```
rushtong: Some more instances of global urls that could be constants in the component instead of being passed in props.

grails-app/controllers/org/broadinstitute/orsp/UserController.groovy
LeoForconesi: This is now changed. Replaced this render error message to `"An error has ocurred trying to update roles"`

grails-app/services/org/broadinstitute/orsp/UserService.groovy
LeoForconesi: Sounds good! I used the message that you suggested.

src/main/webapp/components/Table.js
LeoForconesi: If these callbacks are received as undefined, Bootstrap Table will use their default implementation. So if we do not  pass these props, they will be undefined, and the table will use its default functionality.

src/main/webapp/components/TablePaginator.js
LeoForconesi: Actually we don't need `e`, so I've just removed them. And also I have changed method's calls to a more consistent syntax.

src/main/webapp/projectReview/index.js
LeoForconesi: Indeed. I removed these constants. I belive that these indexes can be removed from react and wepack.config, but I would rather do this change in my next Pr.

PR #181, opened 2019-06-18
[BTRX-701] Fix Search Criteria

tadeork:
## Addresses
This PR takes two stories https://broadinstitute.atlassian.net/browse/BTRX-701 and https://broadinstitute.atlassian.net/browse/BTRX-642
We needed to fix first the search response times to be sure that we're testing the statuses properly.

## Changes
-  Added index in `fundings` table.
-  Added new statuses
-  Now search in two columns of `issue` for status `status` and `approval_status`
-  Label in status form search

## Testing

---


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -728,7 +729,46 @@ class QueryService implements Status {
...
+        final String query = "SELECT i.id id, " +
+                "i.project_key projectKey, " +
+                "i.type type, " +
+                "i.status status, " +
+                "i.summary summary, " +
+                "i.reporter reporter, " +
+                "i.update_date updated, " +
+                "i.expiration_date expirationDate " +
+                "FROM issue i " +
+                "WHERE i.id IN (" + issueIds.join(",") + ")"
```
rushtong: Is there a limit to the number of ids here? I know that oracle has a 1000 item limit, not sure if mysql or cloudsql has the same kind of thing.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -728,7 +729,46 @@ class QueryService implements Status {
...
+            issueSearchItem.setStatus(it.get("status").toString())
+            issueSearchItem.setSummary(it.get("summary").toString())
+            issueSearchItem.setReporter(it.get("reporter").toString())
+            if (it.get("expirationDate") != null) {
+                issueSearchItem.setExpirationDate(it.get("expirationDate"))
+            }
+            if (it.get("updated") != null) {
+                issueSearchItem.setUpdateDate(it.get("updated"))
+            }
+            result.push(issueSearchItem)
```
rushtong: Curious about two things. 
1. Why this is an ordered list? The query is not ordered, but you are forcing the dtos into an array list, so why not just a set?
2. Why use `push`? That's a groovy add-on that just does `add` under the hood - you might have better performance with `add`

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -728,7 +729,46 @@ class QueryService implements Status {
...
+                "i.reporter reporter, " +
+                "i.update_date updated, " +
+                "i.expiration_date expirationDate " +
+                "FROM issue i " +
+                "WHERE i.id IN (" + issueIds.join(",") + ")"
+
+        List<Issue> result = new ArrayList<Issue>()
+
+        getSqlConnection().rows(query).each{
+            Issue issueSearchItem = new Issue()
```
rushtong: Why create a new issue here? You could let hibernate give you a populated `Issue` object from the query which would avoid the awkward parsing, e.g. https://github.com/broadinstitute/orsp-pub/blob/feature/grails-app/services/org/broadinstitute/orsp/QueryService.groovy#L314

This is also a risk because it introduces a new, untested, pattern for hydrating objects. I do worry about what Hibernate will do with all of these dirty (unsaved) new objects and whether that introduces a separate set of performance/memory concerns.

tadeork: If I do as the example the search performance drops drastically from seconds up to five minutes. On the other hand we're not doing anything with the Issue objects but showing the results and only the fields that are mapped, there's no need of other info. So we're not using unnecessary memory.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
tadeork: MySQL has 1GB of limit https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet
I think this will not be a problema.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
rushtong: The concern isn't that we are doing anything with unnecessary memory, the concern is that Grails/Hibernate is doing something. `Issue` is a domain object which means Grails treats it differently than other classes. It could be the case that we're creating memory leaks this way, but I just don't know for sure which is why I bring up the question.

Another alternative is to create some other kind of object that is not a domain object and pass that along as the list of things to return as search results.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
rushtong: This was my question, so I think there is no real limit to the number of elements in the in-clause: https://stackoverflow.com/questions/4275640/mysql-in-condition-limit/4275704

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
tadeork: I don't quite get the concern. The issue objects are read only in this scenario. Another thing, I already try to use a DTO and got very complex very fast. But when I added the index I found that I there was no need to use a DTO because it was very fast and got a lot simpler. Also I found that I was going to have to change the way is used `PermissionService.issueIsForbiden`
In this commit is more or less that approach 87cba6a

rushtong: This is also one of my concerns: 

> The issue objects are read only in this scenario.

This method assumes that these are read-only, but there is nothing to guarantee that. Imagine if some controller called this method, got a bunch of issues back, and tried to do stuff with them. You'd now have modified `Issue` objects that could be re-saved, and only partially (meaning data will be deleted) because you didn't fully populate them from this method. You've chosen a non-standard way to use Grails domain classes which could lead to real problems.

tadeork:
I've made several changes. I think we need another round of testing from @wloforteBroad to be sure that everything is OK.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,61 @@ class PermissionService implements UserInfo {
     UserService userService
 
     // get issue's collaborators as a List<String>
-    List<String> getIssueCollaborators(Issue issue) {
-      issue.extraProperties.findAll ({ it.name == IssueExtraProperty.COLLABORATOR }).collect { property -> property.value }
+    List<String> getIssueCollaborators(Map<String, List<String>> extraProperties) {
+        List<String> results = new ArrayList<String>()
+
+        def collaboratorsValues = extraProperties.findAll ({ it.key == IssueExtraProperty.COLLABORATOR }).values()
```
rushtong: nit: avoid `def` when possible. This is one of those times 😄

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,61 @@ class PermissionService implements UserInfo {
...
+        if (collaboratorsValues.size() != 0) {
+            results = collaboratorsValues.first()
+        }
+
+        collaboratorsValues = extraProperties.findAll ({ it.key == IssueExtraProperty.COLLABORATORS }).values()
+        if (results.size() == 0 && collaboratorsValues.size() != 0) {
+            results = collaboratorsValues.first()
+        }
+
+        results
```
rushtong: This whole method is a little confusing. 
* Why aren't the collaborators *all* of the collaborators? 
* Why is it only the first one (in each of the two current cases)? 
* Why do we mutate the results? 
* Why iterate twice over the `extraProperties`? 

It would make more sense to me to have something like this:
```
List<String> getIssueCollaborators(Map<String, List<String>> extraProperties) {
    extraProperties.
        findAll({ [IssueExtraProperty.COLLABORATOR, IssueExtraProperty.COLLABORATORS].contains(it.key) }).
        values()
}
```

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,61 @@ class PermissionService implements UserInfo {
...
     // get issue's pms as a List<String>
-    List<String> getIssuePMs(Issue issue) {
-      issue.extraProperties.findAll ({ it.name == IssueExtraProperty.PM }).collect { property -> property.value }
+    List<String> getIssuePMs(Map<String, List<String>> extraProperties) {
+        List<String> results = new ArrayList<String>()
+        def collaboratorsValues = extraProperties.findAll ({ it.key == IssueExtraProperty.PM }).values()
+        if (collaboratorsValues.size() != 0) {
+            results = collaboratorsValues.first()
+        }
+        results
```
rushtong: Here again, why are we only showing a single one? The method name, return type, and comments indicate very different behavior.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,61 @@ class PermissionService implements UserInfo {
...
     // get issue's pis as a List<String>
-    List<String> getIssuePIs(Issue issue) {
-      issue.extraProperties.findAll ({ it.name == IssueExtraProperty.PI }).collect { property -> property.value }
+    List<String> getIssuePIs(Map<String, List<String>> extraProperties) {
+        List<String> results = new ArrayList<String>()
+        def collaboratorsValues = extraProperties.findAll ({ it.key == IssueExtraProperty.PI }).values()
+        if (collaboratorsValues.size() != 0) {
+            results = collaboratorsValues.first()
+        }
+        results
```
rushtong: More of the same feedback.

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,61 @@ class PermissionService implements UserInfo {
...
+    def issueIsForbidden(IssueSearchItemDTO issue, String userName, boolean isAdmin, boolean isViewer) {
+        Map<String, List<String>> extraProperties = issue.extraProperties
+        userHasIssueAccess(issue.reporter, extraProperties, userName, isAdmin, isViewer)
+    }
+
+    def userHasIssueAccess(String reporter, Map<String, List<String>> extraProperties, String userName, boolean isAdmin, boolean isViewer) {
+        boolean userHasAccess = (reporter == userName
+                || getIssueCollaborators(extraProperties).indexOf(userName) >= 0
+                || getIssuePMs(extraProperties).indexOf(userName) >= 0
+                || getIssuePIs(extraProperties).indexOf(userName) >= 0
```
rushtong: These boolean checks should be looking at *all* of the PMs, PIs, Collaborators, etc. IndexOf is probably the wrong approach here. I think you want to check `contains` on the list.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                "WHERE i.id IN (" + issueIds.join(",") + ")"
+
+        Set<IssueSearchItemDTO> resultDTO = new HashSet<IssueSearchItemDTO>()
+        IssueSearchItemDTO issueSearchItemDTO
+        String currentProjectKey = ""
+
+        getSqlConnection().rows(query).each {
+            if (it.get("projectKey") == currentProjectKey) {
+                if (it.get("type") != IssueType.CONSENT_GROUP.name) {
+                    def extraProp = issueSearchItemDTO.extraProperties.get(it.get("name").toString()) ?
```
rushtong: avoid `def` here.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                }
+                if (it.get("updated") != null) {
+                    issueSearchItemDTO.setUpdateDate(it.get("updated"))
+                }
+                if (it.get("type") != IssueType.CONSENT_GROUP.name) {
+                    def extraProp = issueSearchItemDTO.extraProperties.size() != 0 ? issueSearchItemDTO.extraProperties.get(it.get("name").toString()) : []
+                    extraProp.push(it.get("value").toString())
+                    issueSearchItemDTO.extraProperties.put(it.get("name").toString(), extraProp)
+                }
+            }
```
rushtong: Pretty much *all* of this logic should be in a constructor for `IssueSearchItemDTO` that understands how to construct itself from a sql query row.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                    extraProp.push(it.get("value").toString())
+                }
+            } else {
+                if (currentProjectKey != "") {
+                    resultDTO.add(issueSearchItemDTO)
+                }
+                currentProjectKey = it.get("projectKey")
+                issueSearchItemDTO = new IssueSearchItemDTO()
+
+                issueSearchItemDTO.setId((Integer)it.get("id"))
```
rushtong: I'm concerned about the casting here. We should strive mightily to avoid this. If you put this logic into the constructor, you could use a try block around a `Integer.valueOf()` line.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                issueSearchItemDTO.setProjectKey(it.get("projectKey").toString())
+                issueSearchItemDTO.setType(it.get("type").toString())
+                issueSearchItemDTO.setStatus(it.get("status").toString())
+                issueSearchItemDTO.setSummary(it.get("summary").toString())
+                issueSearchItemDTO.setReporter(it.get("reporter").toString())
+                if (it.get("expirationDate") != null) {
+                    issueSearchItemDTO.setExpirationDate(it.get("expirationDate"))
+                }
```
rushtong: Why the if condition? Could be a one liner with safe-null checks:
```
issueSearchItemDTO.setExpirationDate(it.get("expirationDate")?.toString())
```

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                if (it.get("updated") != null) {
+                    issueSearchItemDTO.setUpdateDate(it.get("updated"))
+                }
+                if (it.get("type") != IssueType.CONSENT_GROUP.name) {
+                    def extraProp = issueSearchItemDTO.extraProperties.size() != 0 ? issueSearchItemDTO.extraProperties.get(it.get("name").toString()) : []
```
rushtong: Avoid def.

rushtong: Confused by this ... extraProp is either a single string value or an empty list?

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,73 @@ class QueryService implements Status {
...
+                    extraProp.push(it.get("value").toString())
+                    issueSearchItemDTO.extraProperties.put(it.get("name").toString(), extraProp)
```
rushtong: I'm finding this logic confusing ... can you comment on what's happening here?

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
tadeork: It's not the first of the values but the result of this is a array with the Collaborators user names within an array. I need to take out from there the actual values. This suggestion is working but I still need use `.first()`

rushtong: If that's true, then there's something wrong with how you're constructing the values inside `collaboratorsValues`. It should just be just a simple list of names. If instead it is a list of lists, then the underlying method should be super clear about what it's returning.

tadeork: I not agree, `Issue`s are being instantiated in some way that the `extraProperties` are in the same way as my DTO. But I'll fix boths.

vvicario:
@rushtong Can you please take a look and let us know if we need to do something else? Thanks!

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,65 @@ class QueryService implements Status {
...
+                "i.type type, " +
+                "i.status status, " +
+                "i.summary summary, " +
+                "i.reporter reporter, " +
+                "i.update_date updated, " +
+                "i.expiration_date expirationDate, " +
+                "iep.* " +
+                "FROM issue i LEFT JOIN issue_extra_property iep " +
+                "ON (iep.project_key = i.project_key AND iep.name in ('pm','pi','collaborator')) " +
+                "WHERE i.id IN (" + issueIds.join(",") + ")"
```
rushtong: This line is subject to sql injection attacks. This should be a named argument. There are lots of other examples in this class that do this.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,14 +727,65 @@ class QueryService implements Status {
...
+
+        Set<IssueSearchItemDTO> resultDTO = new HashSet<IssueSearchItemDTO>()
+        IssueSearchItemDTO issueSearchItemDTO
+        String currentProjectKey = ""
+
+        getSqlConnection().rows(query).each {
+            if (it.get("projectKey") == currentProjectKey) {
```
rushtong: All of this logic should be in a constructor for `IssueSearchItemDTO`. That class should accept a `GroovyRowResult` as an argument (that's what `it` is in this line) and perform all of the logic there. Having this in the query service breaks encapsulation and makes it impossible to re-use `IssueSearchItemDTO` in any other context.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
rushtong: Any feedback on this question? This looks confusing enough to warrant some kind of comment that describes what it's doing.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
rushtong: @vvicario - why was this marked as resolved? I still think this comment should be addressed.

grails-app/utils/org/broadinstitute/orsp/utils/IssueUtils.groovy
```diff
@@ -105,25 +106,20 @@ final class IssueUtils {
         null
     }
 
-    static Map<String, Object> generateArgumentsForRedirect(Issue issue, String id, String tab) {
+    static Map<String, Object> generateArgumentsForRedirect(String type, String id, String tab) {
         Map<String, Object> arguments = new HashMap<>()
-        if (isConsentGroup(issue)) {
+        if (type == IssueType.CONSENT_GROUP.name) {
```
rushtong: I don't understand why this change (in combination with removing `isConsentGroup()`) was necessary. Are there issues with the wrong controller? Also, this change seems very out of scope for the task of improving search performance.

grails-app/utils/org/broadinstitute/orsp/utils/IssueUtils.groovy
```diff
@@ -105,25 +106,20 @@ final class IssueUtils {
...
+            tab != null ? arguments.put("params", [projectKey: id, tab: tab]) :
+                    arguments.put("params", [projectKey: id])
         }
         arguments
     }
-
-    static boolean isConsentGroup(Issue issue) {
-        return issue.type == IssueType.CONSENT_GROUP.name || issue.controller == IssueType.CONSENT_GROUP.name
-    }
-
```
rushtong: See comment above.

grails-app/utils/org/broadinstitute/orsp/utils/IssueUtils.groovy
tadeork: It seems out of scope but I found out that it was simpler to use this method with a String for the IssueType than sending the Issue. And also, and because I couldn't use this method with IssueSearchItemDTO. While I was figuring out how to use this method with Issue Search DTO I found that this approach was simpler, solve my problem and needed less code.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
tadeork: Agree.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,15 +727,58 @@ class QueryService implements Status {
...
+                if (currentProjectKey != "") {
+                    resultDTO.add(issueSearchItemDTO)
+                }
+
+                currentProjectKey = it.get("projectKey")
+                issueSearchItemDTO = new IssueSearchItemDTO(it.toSorted())
 
+                if (it.get("type") != IssueType.CONSENT_GROUP.name) {
+                    issueSearchItemDTO.setExtraProperty(it.get("name").toString(), it.get("value").toString())
+                }
```
rushtong: Note that lines 763-765 are exactly the same as 774-776. These, too, should be in the constructor and not here in the query service. In fact, I would suggest that the entire if-else should be there. If you have custom constructor logic spread out in different places, you lose the ability to rely on the class's ability to create itself outside of this query service context. All knowledge about what an `IssueSearchItemDTO` is should be in the constructor.

tadeork:
This last approach works well but a little slower than the first intent. Any way we think we can improve in the future with pagination (we've to create a ticket to track that) as we did with Fundings, Roles Management using Bootstrap Table 2.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -725,15 +726,31 @@ class QueryService implements Status {
...
+        final String query = "SELECT * FROM issue i WHERE i.id IN (:issueIds)"
+        SessionFactory sessionFactory = grailsApplication.getMainContext().getBean('sessionFactory')
+        final session = sessionFactory.currentSession
+        final SQLQuery sqlQuery = session.createSQLQuery(query)
+        final results = sqlQuery.with {
+            addEntity(Issue)
+            setParameterList('issueIds', issueIds)
+            list()
+        }
+        results
```
rushtong: @tadeork - you are correct here, this will be a little slower than processing the raw `GroovyRowResults` directly. You don't *have* to use hibernate here (you are with the `addEntity` call). I don't know the exact syntax off-hand, but it might be something as simple as 
```
        final results = sqlQuery.with {
            setParameterList('issueIds', issueIds)
            list()
        }
```
And that should give you a list of raw row results without having to cast each one as an `Issue` object (which happens under the hood by GORM/hibernate).

grails-app/services/org/broadinstitute/orsp/PermissionService.groovy
```diff
@@ -10,31 +10,45 @@ class PermissionService implements UserInfo {
...
     }
 
     // verifies if logged user belongs to some user list ....
-    def issueIsForbidden(Issue issue, String userName, boolean isAdmin, boolean isViewer) {
+    Boolean issueIsForbidden(Issue issue, String userName, boolean isAdmin, boolean isViewer) {
+        Map<String, List<String>> extraProperties = issue.extraPropertiesMap
+        userHasIssueAccess(issue.reporter, extraProperties, userName, isAdmin, isViewer)
+    }
+
+    Boolean issueIsForbidden(IssueSearchItemDTO issue, String userName, boolean isAdmin, boolean isViewer) {
```
rushtong: Are you even using this any more now that your query returns full `Issue` objects?

src/main/groovy/org/broadinstitute/orsp/IssueSearchItemDTO.groovy
```diff
@@ -0,0 +1,48 @@
+package org.broadinstitute.orsp
+
+import groovy.util.logging.Slf4j
+
+@Slf4j
+class IssueSearchItemDTO {
```
rushtong: This looks unused now.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
tadeork: This is not working. I need some how to cast the results to Issues.

rushtong: Something like this should also work - it doesn't use hibernate at all (sqlQuery still uses hibernate to generate a raw sql query ... fast, but not as fast as groovy sql)

```
getSqlConnection().rows(query, ["issueIds": issueIds]).collect { it -> ... make something from the it, which is a GroovyRowResult ... }
```

tadeork: I already tried this approach and also didn't work. It doesn't returns any result. I tried several approaches and it was frustrating bcs only a few ones work but slowly.

rushtong: Yeah, I was running a local instance to test and it does look like that doesn't work. No worries on this topic then.

PR #182, opened 2019-06-18
BTRX-694 OSAP Integration

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-694
---


src/main/groovy/org/broadinstitute/orsp/IssueInfo.groovy
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.orsp
+
+class IssueInfo {
```
rushtong: Nit: naming - this is very specific to what OSAP is looking for, I would suggest a more OSAP specific name.

vvicario: Agree. could it be OSAPDataFeed ?

rushtong: Works for me 👍

PR #184, opened 2019-06-18
[BTRX-710] Permission functionality

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-710

## Changes
Redirects to home if user has no access to CG or Project.

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/ConsentGroupController.groovy
```diff
@@ -183,7 +183,11 @@ class ConsentGroupController extends AuthenticatedController {
     @Override
     show() {
         Issue issue = queryService.findByKey(params.id)
-        if (issueIsForbidden(issue)) {
+        Boolean isUserReporter = false
+        if (params.projectKey != null) {
+            isUserReporter = queryService.findByKey(params.projectKey).reporter == getUser().userName
```
rushtong: Why make the same query twice? On line 185, you already have the issue. This should work:
```
Boolean isUserReporter = issue?.reporter == getUser().userName
```

tadeork: Yes but there are not the same issue. If the user is trying to get into a CG from a Project I need to know if he/she has access. I need not only the id of the CG but also the id of the Project to know if the user is the reporter. One query for the Project and the other one for the CG.

rushtong: Ah ... yes, I missed that. You're querying on two things, the consent group which is `params.id` and then the project, which is `params.projectKey`. So, my feedback here is not accurate, thank you!

One more detail, `queryService.findByKey(params.projectKey).reporter` could be safer if you did this: `queryService.findByKey(params.projectKey)?.reporter`

wloforteBroad:
> Not sure this solves the problem of the user searching from the nav bar:
> 
> > NOTE: Also fix the following: If a user has not been granted access to a project, and searches for the project in the search page, the system successfully shows the user that the projects exists, but does not allow access.  HOWEVER, if the user searchers in the search field located in the banner, the system provides full access to the record that she should not have access to.
> 
> That code is in `SearchController.getMatchingIssues()` and `_topNav.gsp`. If I understand how this works, a user would click on an unallowed entry in the top nav search bar, then get redirected to the main page with no information as to why or what happened. At a minimum, we should be telling the user something about why they can't get to where they want to go.
> 
> I think this story needs some refinement, and this solution needs to be run by ORSP for approval.

This was already happening, I totally agree with you and I'll create a story to show a message to the user when they have no permission to see an issue. We were discussing that same thing with Jaime yesterday.

rushtong:
@wloforteBroad  - Thank you, yes, that was happening, but it was accidental and a side-effect of the other role-based changes. With this PR, a user on the search page (or in any other part of the site) will be redirected and therefore lose the context of where they were. 

I think a better solution were to prevent the user from following the link in the first place. Imagine that the autocomplete returned the same results now, but with no link for the ones the user doesn't have access to. Then, the autocomplete widget could check for that condition and show something nice in the UI, i.e., treat it differently and make it not clickable with some kind of information saying "Hey, you can't go there!". Better to capture the error before the user clicks rather than after since we know, up front, what the user can and cannot do.

wloforteBroad:
@rushtong Yes, totally agree. But do you think we can do it as part of a new story, or do it on this one?

rushtong:
@wloforteBroad - I don't have a strong opinion on that question, I think it's a PO decision on whether the functionality here meets the requirements. If Andrea is OK with this partial solution, then yes, move forward and create a new story for what this should be doing.

wloforteBroad:
@rushtong this story was to improve a couple of thing that were not working as the ORSP team wanted, so I think is safe to resolve this in a different story. The two issues to solve were:
1- Users with permissions to see a Project didn't have access to related Consent groups, they can access now.
2- User without access to a Project or CG could access them by searching in the Header and clicking them, they are not able to do that any more.

rushtong:
@wloforteBroad 

> 1- Users with permissions to see a Project didn't have access to related Consent groups, they can access now.

Agreed. 

> 2- User without access to a Project or CG could access them by searching in the Header and clicking them, they are not able to do that any more.

Agreed ... but the current PR implementation adds a new bug, i.e. a user can be redirected to a page they did not want to go to with no explanation. I approved this PR - it's up to Andrea and team to approve the implementation of the story.

PR #187, opened 2019-06-24
BTRX-678 Implement Project - Sample/Data Cohort link approval process

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-678

## Testing

---


grails-app/services/org/broadinstitute/orsp/PersistenceService.groovy
```diff
@@ -53,4 +53,10 @@ class PersistenceService {
         consentCollectionLink?.delete(flush: true)
     }
 
+    void updateCollectionLinkStatus(String consentKey, String projectKey, String status) {
+        ConsentCollectionLink.executeUpdate("update ConsentCollectionLink ccl set ccl.status = :status where ccl.projectKey = :projectKey and ccl.consentKey = :consentKey",
```
rushtong: nit: This is HQL, not sql, and will work fine. I'm not requesting that you change this, but wanted to point out that this is a different pattern than what we've been striving for, i.e. moving toward using sql directly in persistence services. In general, anything that is GORM or hql has not been nearly as performant as raw sql. It's trivial to write these updates that way, but when we use hql/gorm, it goes through a significant layer of ORM manipulation overhead. For simple updates like this, it's likely that you won't see the impact. In larger updates/queries, you will definitely see a performance impact.

grails-app/views/consentGroup/_confirmationDialog.gsp
```diff
@@ -0,0 +1,42 @@
...
+                      </a>
+                    </g:if>
+                    <g:if test="${actionKey == 'reject' || actionKey == 'unlink'}">
+                      <a class="btn btn-primary" href="${createLink(controller: 'consentGroup', action: 'breakLink', params: [projectKey: issue.projectKey, consentKey: consent.projectKey, type: actionKey])}">
+                        Accept
+                      </a>
+                    </g:if>
+            </div>
+    </div>
+</div>
```
rushtong: Why aren't these react components?

grails-app/views/consentGroup/_requestClarification.gsp
```diff
@@ -0,0 +1,59 @@
...
+    z-index: 1500;
+    overflow-y: scroll;
+    overflow-x: hidden;
+    height: 300px;
+}
+
+.ui-dialog.ui-widget.ui-corner-all.ui-front {
+    overflow: visible;
+}
+</style>
```
rushtong: Same concern here ... we're increasing the tech-debt backlog with this PR.

src/main/webapp/components/Btn.css
```diff
@@ -27,3 +27,114 @@
...
+.cta-container .btn:active,
+.cta-container .btn:focus {
+  outline: none;
+}
+
+.request-clarification {
+  display: inline-block;
+  position: relative;
+  height: 50px;
+  background: url('../../../../grails-app/assets/images/icon-live_help.svg') no-repeat center center;
```
rushtong: Do you know if this gets translated correctly with webpack? I would think that `/assets/...` should work here.

src/main/webapp/components/Btn.css
```diff
@@ -27,3 +27,114 @@
...
+  text-decoration: underline;
+}
+
+.consent-group-panel-body {
+  min-width: 100%;
+}
+
+p.modal-large {
+  font-size: 16px;
+}
```
rushtong: nit: newline

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -61,6 +62,69 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
       }).parent().removeClass("ui-widget-content");
       $(".ui-dialog-titlebar").hide();
     });
+
+    $(".confirmationModal").on('click', function () {
+      $("#confirmation-modal-dialog").load(
+        component.serverURL + "/api/consent-group/confirmation-modal?"
```
rushtong: Ah ... I think I see why you added the gsp. I don't agree with this approach ... all new API calls should be returning raw json objects, not gsps. All html view content should be a react component.

vvicario:
@rushtong I worked on your feedback. Can you please take a look? Thank you

grails-app/services/org/broadinstitute/orsp/PersistenceService.groovy
vvicario: :+1: I moved to sql

grails-app/views/consentGroup/_requestClarification.gsp
vvicario: Agree, I removed it and I moved to react components

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -199,4 +200,22 @@ class NewConsentGroupController extends AuthenticatedController {
...
+    def approveLink () {
+        try {
+            queryService.updateCollectionLinkStatus(params.consentKey, params.projectKey, IssueStatus.Approved.name)
+            response.status = 200
+        } catch (Exception e) {
+            response.status = 500
+            log.error("Exception updating collection links status: " + e)
+            flash.error = "Error updating collection links status: " + e
+        }
+        redirect(controller: "project", action: "main", params: [projectKey: params.projectKey, tab: "consent-groups"])
```
rushtong: I don't think we should be redirecting here (or anywhere, for that matter). If we redirect, that means we lose all context and state from the page that this was called from. The way our APIs should work is that they perform some kind of function, then return the result of that function to the React component that made the call. The component then reads that information and performs state updates as needed. We're trying to get away from the multi-page pattern and moving toward the single-page pattern.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -401,6 +401,17 @@ class QueryService implements Status {
...
+    void updateCollectionLinkStatus(String consentKey, String projectKey, String status) {
+        final session = sessionFactory.currentSession
+        final String query =
+                ' update consent_collection_link set status = :status ' +
+                'where project_key = :projectKey and consent_key = :consentKey'
+        final sqlQuery = session.createSQLQuery(query)
+        sqlQuery.setParameter('projectKey', projectKey)
+        sqlQuery.setParameter('consentKey', consentKey)
+        sqlQuery.executeUpdate()
+    }
```
rushtong: Would it make more sense to return something useful to the caller? That way any upstream APIs can return the result of the update to the calling page.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
rushtong: I think I see why you're doing this ... and it has to do with the fact that the caller is in a GSP (see  `grails-app/views/consentGroup/list.gsp`). I'm not sure what the best tactical approach is here - maybe that page needs to go through the react conversion first, or we can add this bit of tech debt to handle in whatever story does handle this.

vvicario: I agree, when I moved the component to react I forgot to remove the redirection. I'll do that

grails-app/controllers/org/broadinstitute/orsp/ConsentGroupController.groovy
```diff
@@ -126,10 +126,18 @@ class ConsentGroupController extends AuthenticatedController {
...
-        def issue = queryService.findByKey(params.projectKey)
-        def links = ConsentCollectionLink.findAllByProjectKeyAndConsentKey(params.projectKey, params.consentKey)
-        deleteCollectionLinks(links)
-        redirect(controller: "project", action: "main", params: [projectKey: params.projectKey, tab: "consent-groups"])
+        try {
+            queryService.updateCollectionLinkStatus(params.consentKey, params.projectKey, params.type == "unlink" ? IssueStatus.Unlinked.name : IssueStatus.Rejected.name)
+            List<ConsentCollectionLink> links = ConsentCollectionLink.findAllByProjectKeyAndConsentKey(params.projectKey, params.consentKey)
+            deleteCollectionLinks(links)
+            response.status = 200
+            render([message: "Links are been deleted"] as JSON)
```
rushtong: nit: It would be a lot nicer to describe for the user what links were deleted. As it stands now, this is a pretty opaque operation.

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -50,6 +50,9 @@ class UrlMappings {
         '/api/consent-groups'(controller: 'consentGroup', action: 'projectConsentGroups', method: 'GET')
         '/api/consent-group/upload-modal'(controller: 'consentGroup', action: 'loadModalWindow', method: 'GET')
 
+        '/api/break-link'(controller: 'consentGroup', action: 'breakLink', method: 'DELETE')
```
rushtong: Is the 'DELETE' accurate here? You're reading params which means there is a post-body content. This seems more like a POST than a DELETE.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -199,4 +200,28 @@ class NewConsentGroupController extends AuthenticatedController {
...
+              response.status = 400
+              render([message: 'Error updating collection links, please check specified parameters.'] as JSON)
+            } else {
+              response.status = 200
+              render([message: 'Collection links are been updated.'] as JSON)
+            }
+        } catch (Exception e) {
+            response.status = 500
+            log.error("Exception updating collection links status: " + e)
+            flash.error = "Error updating collection links status: " + e
```
rushtong: Flash errors in Grails are part of the user's session object. GSPs have the ability to read that from the session, so in this context, this really doesn't have any effect.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -199,4 +200,28 @@ class NewConsentGroupController extends AuthenticatedController {
...
+     */
+    def approveLink () {
+        try {
+            boolean isUpdated = queryService.updateCollectionLinkStatus(params.consentKey, params.projectKey, IssueStatus.Approved.name)
+            if (!isUpdated) {
+              response.status = 400
+              render([message: 'Error updating collection links, please check specified parameters.'] as JSON)
+            } else {
+              response.status = 200
+              render([message: 'Collection links are been updated.'] as JSON)
```
rushtong: Similar concern with respect to returning information to the user. A json list of created links would be very appropriate.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -567,19 +580,29 @@ class QueryService implements Status {
         issue
     }
 
+    Boolean areLinksApproved(String projectKey, String consentKey) {
+        List<ConsentCollectionLink> links = ConsentCollectionLink.findAll("from ConsentCollectionLink as ccl where ccl.projectKey= :projectKey and ccl.consentKey = :consentKey",
```
rushtong: I don't recommend GORM style queries. This can easily be a `getSqlConnection()` call that just queries for the status.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -567,19 +580,29 @@ class QueryService implements Status {
...
+                [consentKey: consentKey, projectKey: projectKey])
+        ArrayList approvedLinks = links.findAll {
```
rushtong: In general, favor List over ArrayList. Also, please use strong typing, i.e. `List<ConsentCollectionLink>`. I have a feeling that once you write a more concise query to get all approved statuses from the ccl table, you won't need to do this at all.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -567,19 +580,29 @@ class QueryService implements Status {
...
+    }
+
     /**
      * Find issues by keys
      *
      * @param keys The issue keys
      * @return List of Issues that match the query
      */
-    Collection<Issue> findByKeys(Collection<String> keys) {
+    Collection<Issue> findByKeys( Map<String, ConsentCollectionLink> keys) {
```
rushtong: nit spacing:
```suggestion
    Collection<Issue> findByKeys(Map<String, ConsentCollectionLink> keys) {
```

grails-app/views/consentGroup/list.gsp
```diff
@@ -1,127 +1,164 @@
...
-        var url = new URLSearchParams(window.location.search);
-        if (url.get('tab') === 'consent-groups' && url.has('new')) {
-            $('#alert').fadeIn('slow', function(){
-                   $('#alert').delay(8000).fadeOut();
-                   history.pushState({}, null, window.location.href.split('&')[0]);
-            });
-        }
-    });
+  // Display for 8 seconds a message indicating the submission of a new consent group. This is temporary until this page is moved to react.
+  $(document).ready(function () {
```
rushtong: nit spacing:
```suggestion
  $(document).ready(function() {
```

grails-app/views/consentGroup/list.gsp
```diff
@@ -1,127 +1,164 @@
...
+    var url = new URLSearchParams(window.location.search);
+    if (url.get('tab') === 'consent-groups' && url.has('new')) {
+      $('#alert').fadeIn('slow', function () {
```
rushtong: nit spacing:
```suggestion
      $('#alert').fadeIn('slow', function() {
```

vvicario:
@rushtong Can you take a look at this please? Thank you!

grails-app/controllers/org/broadinstitute/orsp/ConsentGroupController.groovy
```diff
@@ -126,10 +125,19 @@ class ConsentGroupController extends AuthenticatedController {
...
-        def links = ConsentCollectionLink.findAllByProjectKeyAndConsentKey(params.projectKey, params.consentKey)
-        deleteCollectionLinks(links)
-        redirect(controller: "project", action: "main", params: [projectKey: params.projectKey, tab: "consent-groups"])
+        try {
+            queryService.updateCollectionLinkStatus(params.consentKey, params.projectKey, params.type == "unlink" ? CollectionLinkStatus.Unlinked.name : CollectionLinkStatus.Rejected.name)
+            List<ConsentCollectionLink> links = queryService.findConsentCollectionLinksByProjectKeyAndConsentKey(params.projectKey, params.consentKey)
+            deleteCollectionLinks(links)
+            response.status = 200
+            render([message: "Links with the following ids: " + links.collect{it.sampleCollectionId}.join(",")  +
+                    " for the specified consent key: " + params.consentKey + " and project key " + params.projectKey + " are been deleted"] as JSON)
```
rushtong: * nit, language: `are been deleted` -> `have been deleted`
* sampleCollectionId can be null so you might have empty list elements in `links.collect{it.sampleCollectionId}.join(",")`

grails-app/controllers/org/broadinstitute/orsp/ConsentGroupController.groovy
```diff
@@ -126,10 +125,19 @@ class ConsentGroupController extends AuthenticatedController {
...
+        } catch (Exception e) {
+            response.status = 500
+            log.error("Exception deleting collection links: " + e)
+            render([error: "There was an error trying to unlink collection links"] as JSON)
```
rushtong: nit language (but maybe confirm with Andrea's team on this too): 
```suggestion
            render([error: "There was an error trying to remove project sample data cohort associations."] as JSON)
```

grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -60,4 +60,28 @@ class ClarificationController extends AuthenticatedController {
         }
         response
     }
+
+    def collectionRequestClarification() {
+        Issue issue = queryService.findByKey(params.id)
+        if (params.comment) {
+            Comment comment = persistenceService.saveComment(issue.projectKey,  getUser()?.displayName, params.comment)
+            String toAddresses = userService.findUser(params.pm)?.collect {it.emailAddress}
+            String fromAddress = (String) getUser()?.emailAddress
```
rushtong: emailAddress is already a string ... why the cast? Remember, casting is *always* a code smell. If you're doing it, you should be 1000% sure that's the only way to do what you need.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -141,6 +141,21 @@ class QueryService implements Status {
         getSqlConnection().rows(query, ["consentKey": consentKey]).collect { it.get("sample_collection_id").toString() }
     }
 
+    /**
+     * Find all sample collection ids that have been associated to the provided consent.
+     *
+     * @return Collection of distinct sample collection ids
```
rushtong: nit: this is a copy/paste error

src/main/groovy/org/broadinstitute/orsp/CollectionLinkStatus.groovy
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.orsp
+
+enum CollectionLinkStatus {
+
+
+    Approved("Approved"),
```
rushtong: Enum instances are constants and should follow the all uppercase naming convention ... (see https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)
```suggestion
    APPROVED("Approved"),
```

src/main/webapp/components/RequestClarificationDialog.js
```diff
@@ -26,20 +32,42 @@ export const RequestClarificationDialog = hh(class RequestClarificationDialog ex
...
+      prev.pm = [{key:''}],
+      prev.submit = false;
       return prev;
     });
     this.props.closeModal();
   };
 
+  validateClarification() {
+    let isValid = false;
+    if(this.props.linkClarification === true) {
```
rushtong: nit: spacing
```suggestion
    if (this.props.linkClarification === true) {
```

PR #188, opened 2019-06-26
BTRX - 718 - Reviewer notification

LeoForconesi:
## Addresses
Please link to Jira ticket here
[BTRX-718](https://broadinstitute.atlassian.net/browse/BTRX-718) : ORSP - Include reviewer as recipient on any Reivew Edit 

## Changes
- Add the edit's user creator when a new edit is updated to project or consent
- If the edit is approved, rejected or a new request clarification is added, the creator of the edit will be notified.
- Fix Redirection bug on top-nav search box for consent groups
- Fix String mapping for security option in consent collection link "uncertain" option

## Testing
- Submit an edit suggestion on a project, then approve this edition. The user who has created the suggestion will receive an Edit Approval e-mail.

---


grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -23,8 +27,12 @@ class ClarificationController extends AuthenticatedController {
                 render([message: "Error saving comment"])
             }
 
+            IssueReview issueReview = issueReviewService.findByProjectKey(issue.projectKey)
             List<String> toAddresses = new ArrayList<>()
             String fromAddress = (String) getUser()?.emailAddress
+            if (issueReview != null) {
+                toAddresses.addAll(userService.findUser(issueReview.getEditCreatorName())?.emailAddress)
```
dags: add single email, no need for addAll.

grails-app/services/org/broadinstitute/orsp/IssueReviewService.groovy
```diff
@@ -1,5 +1,7 @@
 package org.broadinstitute.orsp
 
+import com.google.gson.Gson
+import grails.converters.JSON
```
dags: Removed unused imports

grails-app/services/org/broadinstitute/orsp/IssueReviewService.groovy
```diff
@@ -20,7 +22,8 @@ class IssueReviewService {
             ir.delete(flush: true)
         }
         if (type == 'reject') {
-            notifyService.sendEditsDisapprovedNotification(Issue.findByProjectKey(ir.projectKey))
+            String editCreatorName = ir.getEditCreatorName()
+            notifyService.sendEditsDisapprovedNotification(Issue.findByProjectKey(ir.projectKey), editCreatorName)
```
dags: Could use ir.getEditCreatorName() directly in the request

grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -23,8 +27,12 @@ class ClarificationController extends AuthenticatedController {
                 render([message: "Error saving comment"])
             }
 
+            IssueReview issueReview = issueReviewService.findByProjectKey(issue.projectKey)
             List<String> toAddresses = new ArrayList<>()
             String fromAddress = (String) getUser()?.emailAddress
+            if (issueReview != null) {
+                toAddresses.add(userService.findUser(issueReview.getEditCreatorName())?.emailAddress)
```
rushtong: This could add a null value to the list. I think you'd want to filter those out.

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
```diff
@@ -657,12 +657,14 @@ class NotifyService implements SendgridSupport, Status {
         sendMail(mail, getApiKey(), getSendGridUrl())
     }
 
-    Map<Boolean, String> sendEditsApprovedNotification(Issue issue) {
+    Map<Boolean, String> sendEditsApprovedNotification(Issue issue, String editCreatorName) {
         String type = issue.type.equals(IssueType.CONSENT_GROUP.getName()) ? "Consent Group" : "Project"
+        User editCreator = userService.findUser(editCreatorName)
```
rushtong: Looking through the code path here, `editCreatorName` could be null.

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
```diff
@@ -673,12 +675,14 @@ class NotifyService implements SendgridSupport, Status {
         sendMail(mail, getApiKey(), getSendGridUrl())
     }
 
-    Map<Boolean, String> sendEditsDisapprovedNotification(Issue issue) {
+    Map<Boolean, String> sendEditsDisapprovedNotification(Issue issue, String editCreatorName) {
         String type = issue.type?.equals(IssueType.CONSENT_GROUP.getName()) ? "Consent Group" : "Project"
+        User editCreator = userService.findUser(editCreatorName)
```
rushtong: Ditto here.

grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
LeoForconesi: I solved iit this way: 
https://github.com/broadinstitute/orsp-pub/pull/188/files#diff-8940ab54e457fae2a2f136b13aa6c6d0R32

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
LeoForconesi: Solved in a similar way as clarificationRequest Controller -> 
https://github.com/broadinstitute/orsp-pub/pull/188/files#diff-0401de3e18519430c4c05ee635382da1R666

LeoForconesi:
@greg could you please check this again?

grails-app/services/org/broadinstitute/orsp/NotifyService.groovy
```diff
@@ -673,12 +674,13 @@ class NotifyService implements SendgridSupport, Status {
...
     }
 
-    Map<Boolean, String> sendEditsDisapprovedNotification(Issue issue) {
+    Map<Boolean, String> sendEditsDisapprovedNotification(Issue issue, String editCreatorName) {
         String type = issue.type?.equals(IssueType.CONSENT_GROUP.getName()) ? "Consent Group" : "Project"
         User user = userService.findUser(issue.reporter)
         NotifyArguments arguments =
                 new NotifyArguments(
                         toAddresses: Collections.singletonList(user.emailAddress),
+                        ccAddresses: editCreatorName != null ? Collections.singletonList(userService.findUser(editCreatorName).emailAddress) : [""],
```
rushtong: Note that this only works because we do a lot of validation on recipients here:
```
    private List<String> filterRecipients(Collection<String> recipients) {
        if (recipients?.contains(ORSP_ADDRESS)) {
            recipients -= ORSP_ADDRESS
            recipients += getOrspSpecialRecipients()
        }
        if (recipients != null) {
            recipients.toSet()
            recipients -= null
            recipients.toList()
        } else {
            Collections.emptyList()
        }
    }
```
TOL ... (thinking out loud) ... would it be better to just pass an empty array in the null case instead of an array with an empty string value?

rushtong: The more I ponder this, the more I think passing an empty array makes more semantic sense. Your code should never try to send an email to `""` ... and that's what this code is saying.

LeoForconesi: Makes sense. As you suggested, this condition returns an empty array for null case. I did a local test round and it works good.

PR #192, opened 2019-06-28
Btrx 714 wloforte

wloforteBroad:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-714

## Changes
Added functionality to show the user which project they have access to and which they don't. In the last case we added a tag with the Name of the person to contact for access and prevented clicking. Also made some great Ui improvements.

## Testing
In the top Nav make a couple of searches that display issues with and without access.

---


grails-app/controllers/org/broadinstitute/orsp/SearchController.groovy
```diff
@@ -47,6 +47,10 @@ class SearchController implements UserInfo {
     }
 
     def getMatchingIssues() {
+        def user = getUser()
+        def userName = user?.userName
+        def isAdmin = isAdmin()
+        def isViewer = isViewer()
```
rushtong: Avoid `def` ... please use full objects.

grails-app/controllers/org/broadinstitute/orsp/SearchController.groovy
```diff
@@ -122,7 +130,7 @@ class SearchController implements UserInfo {
 
     def generalReactTablesJsonSearch() {
         def user = getUser()
-        def userName = user.userName
+        def userName = user?.userName
```
rushtong: Not originally your code, true, but please please use objects instead of perpetuating the usage of `def`

wloforteBroad: No problem. Fixed!

grails-app/controllers/org/broadinstitute/orsp/SearchController.groovy
wloforteBroad: Done, thanks!

PR #197, opened 2019-07-01
[BTRX-713] Add PM column to Search Results Table

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-713

## Changes
Add new column to search result showing the user to contact to request access.

## Testing

---


grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -489,4 +489,15 @@ class IssueService implements UserInfo {
...
     }
 
+    /**
+     * Project can have PM ACTOR CREATOR.
+     * CG only have CREATOR.
+     * @param extraProperties
+     * @return
+     */
+    static Collection<String> getPmOrActor(Map<String, List<String>> extraProperties) {
+        Collection<String> userPm = extraProperties.findAll ({ it.key == IssueExtraProperty.PM }).values().flatten()
```
dags: userPm name seems to indicate a single item but this is a Collection. Probably should be named differently.

grails-app/controllers/org/broadinstitute/orsp/SearchController.groovy
```diff
@@ -152,6 +152,7 @@ class SearchController implements UserInfo {
                 options.irbsOfRecord) {
             rows = queryService.findIssues(options).collect {
                 Map<String, Object> arguments = IssueUtils.generateArgumentsForRedirect(it.type, it.projectKey, null)
+                Collection<String> accessContacts = getProjectAccessContact(it.extraPropertiesMap, it.reporter)
```
rushtong: Minor suggestion. Since these values go right to the UI, it might be nice to sort them here. Alpha is probably the safest bet. That way you are generally returning the list in the same order each time and the UI can decide what to do differently if it wants to. The way it is now, the list will be random depending on how it loads.

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -489,4 +489,15 @@ class IssueService implements UserInfo {
         IssueType.valueOfName(issue?.getType())?.prefix?.toLowerCase()
     }
 
+    /**
+     * Project can have PM ACTOR CREATOR.
+     * CG only have CREATOR.
+     * @param extraProperties
+     * @return
```
rushtong: nit: either remove or fill it in. This is a package protected method, probably doesn't need it. I don't feel strongly about javadoc in general. The only exception is for non-trivial methods that do some substantial work. This one is pretty self explanatory.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -748,7 +748,7 @@ class QueryService implements Status {
             params.put('funding', "%" + options.getFundingInstitute() + "%")
         }
         if (options.userName) {
-            def q = ' u.user_name like :userName '
+            def q = ' u.user_name like :userName  or i.reporter like :userName'
```
rushtong: nit: spacing: 
```suggestion
            def q = ' u.user_name like :userName or i.reporter like :userName'
```

PR #198, opened 2019-07-02
BTRX-720: Add question to "Use Existing SDC" flow

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-720

## Testing

---


src/main/webapp/linkWizard/LinkWizard.js
```diff
@@ -229,21 +238,27 @@ export const LinkWizard = hh( class LinkWizard extends Component {
...
     let consentGroup = false;
     let isValid = true;
-
+    let isDateRangeValid = true;
     if (isEmpty(this.state.consentGroup)) {
       consentGroup = true;
       isValid = false;
     }
-
+    if(this.validateDateRange() == false) {
```
dags: add a space in if( ...

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -187,6 +192,27 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
...
+      return prev;
+    }, () => this.props.updateDateRange(this.state.onGoingProcess, this.state.endDate, this.state.startDate));
+  };
+
+  handleChange = (id) => (date) => {
+    this.setState(prev => {
+      prev[id] = date;
+      return prev;
+    }, () => {
+      if (id == 'startDate' || id == 'endDate') {
```
rushtong: This seems brittle, but I don't have a better implementation solution to point you in a better direction. Can you add a comment here that this is tightly coupled to the rendered components?

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -187,6 +192,27 @@ export const SelectSampleConsent = hh(class SelectSampleConsent extends Componen
...
+        this.props.updateDateRange(this.state.onGoingProcess, this.state.endDate, this.state.startDate);
+      } else {
+        this.props.updateForm(this.state.sampleCollection, id);
+      }
```
rushtong: It looks like the only way this is called is via the two `InputFieldDatePicker` references ... is this else condition reachable?

rushtong: One way to avoid having to do these kinds of else checks is to have differently named handlers for the different things that do the handling, that way the same function doesn't have to do double-duty.

vvicario: I agree, I'll create another handler

PR #202, opened 2019-07-03
[BTRX-727] Remove GDPR question

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-727

## Changes
Removed from international cohort wizard and review components GDPR question.

## Testing

---


src/main/webapp/components/InternationalCohorts.js
```diff
@@ -38,7 +38,7 @@ export const InternationalCohorts = hh(class InternationalCohorts extends Compon
       yesOutput: 3,
       noOutput: DETERMINATION.EXIT,
       answer: null,
-      progress: 17,
+      progress: 20,
```
rushtong: nit: not your code, so does not have to be addressed in this PR. This really should be a calculation, not a hard number. Future changes would only need to alter the formula instead of each particular progress level.

tadeork: Agree.

tadeork: https://broadinstitute.atlassian.net/browse/BTRX-730 <- done

PR #205, opened 2019-07-04
Search Refactor to fix 502 Error on Cloud

dags:
## Addresses
This is a fix the the issue (Err 502) verified on Cloud deployment + Ingress Load Balancer. It is an alternative query to NOT use hibernate addEntity in order to avoid overhead that seems to be causing 502 errors.

## Changes
The search query was replaced with a new one, not using AddEntity and therefore using manaul mapping of returned values. Additionally, user names search was included to avoid extra sql requests. 

## Testing

---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -75,8 +75,8 @@ class QueryService implements Status {
     Collection<String> findAllDiseaseAndPopulationRestrictions() {
         final String query =
                 ' select distinct disease_restrictions_string from data_use_restriction_disease_restrictions ' +
-                ' union ' +
-                ' select distinct disease_restrictions_string from data_use_restriction_disease_restrictions '
+                        ' union ' +
+                        ' select distinct disease_restrictions_string from data_use_restriction_disease_restrictions '
         getSqlConnection().rows(query).collect { it.get("disease_restrictions_string").toString() }
     }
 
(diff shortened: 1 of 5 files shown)


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -786,22 +786,87 @@ class QueryService implements Status {
...
-     * Find issues from a list of Ids
-     * @param issueIds
-     * @return
-     */
-    Set<Issue> findIssuesSearchItemsDTO(ArrayList<Integer> issueIds) {
-        final String query = "SELECT * FROM issue i WHERE i.id IN (:issueIds) and i.deleted = 0"
+/**
+ * Find issues and create DTOs to narrow the amount of data used in the search process
+ * @param issueIds
+ * @return
```
rushtong: nit: either fill out the `@return` or remove both `@`'s. I'm not worried about javadocs here, but we should be consistent.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -786,22 +786,87 @@ class QueryService implements Status {
...
+                }
+                currentProjectKey = it.projectKey
+                issueSearchItemDTO = new IssueSearchItemDTO(it.toSorted())
+
+                if (it.type != IssueType.CONSENT_GROUP.name) {
+                    issueSearchItemDTO.setExtraProperty(it.name.toString(), it.value.toString())
+                    issueSearchItemDTO.setAccessContacts(it.name.toString(), it.displayName.toString())
+                }
+            }
+            resultDTO.add(issueSearchItemDTO)
```
rushtong: I think all of this logic should be in the `IssueSearchItemDTO`. There should be a method that takes the collection of results and returns a`Set<IssueSearchItemDTO>`. By putting all of the constructor logic in the query service, you lose the ability for `IssueSearchItemDTO` to have control over how it is created/populated/etc and therefore is of little value outside of this query service method.

src/main/groovy/org/broadinstitute/orsp/IssueSearchItemDTO.groovy
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+/**
+ * Check if already exists extra properties for the specified name, get the values and
+ * add to the existing ones. If not create an empty Set to add the extra prop value
+ * @param name
+ * @param value
+ */
+    void setExtraProperty(String name, String value) {
+        Set<String> extraProp = this.extraProperties.get(name) != null ? this.extraProperties.get(name) : []
```
rushtong: Might be safer if we used strong typing
```suggestion
        Set<String> extraProp = this.extraProperties.get(name) != null ? this.extraProperties.get(name) : new ArrayList<String>()
```

src/main/groovy/org/broadinstitute/orsp/IssueSearchItemDTO.groovy
```diff
@@ -0,0 +1,68 @@
...
+        extraProp.add(value)
+        this.extraProperties.put(name, extraProp)
+    }
+
+    void setAccessContacts(String name, String value) {
+        Set<String> access  = this.contactNames.get(name) != null ? this.contactNames.get(name) : []
```
rushtong: same feedback here.

PR #207, opened 2019-07-08
BTRX- 712 - Convert role Management

LeoForconesi:
## Addresses
[BTRX-712](https://broadinstitute.atlassian.net/browse/BTRX-712): ORSP - Convert Role Management

## Changes
- Remove Roles Management grails dependency.
- Implement React Router to Main Component in order to improve redirects in React.
- Use Application context path from configurations on router browser in order to dynamically change according to the environment the app is working on (dev / test / prod, etc...)
- Remove unnecessary modules export entries.

## Testing
- Navigate to Roles Management from the navigation bar. 
- Access to Projects or Sample Data Cohorts from any link, it should be correctly redirected.
- Start application from another context, e.g Test. previous redirections should still be ok.

---


src/main/webapp/main/main_index.js
```diff
@@ -1,12 +1,21 @@
...
+import { BrowserRouter } from 'react-router-dom';
 
 ReactDOM.render(
   <ErrorHandler>
-    <Main/>
+    <BrowserRouter basename= {component.contextPath}>
+      <Switch>
+        <Route path= {"/user/rolesManagement"} render = {(routeProps) =>  <RolesManagement {...routeProps}/> }/>
+        <Route path= {"/project/main"} render = {(routeProps) => <Main {...routeProps}/> }/>
+        <Route path= {"/newConsentGroup/main"} render = {(routeProps) => <Main {...routeProps}/> }/>
```
rushtong: Any props passsed into `RolesManagement` or `Main` will be lost here, they'll only have access to `routeProps`. It should be something like this so that `routeProps` are merged with component `props`:
```suggestion
        <Route path= {"/newConsentGroup/main"} render = {(routeProps) => <Main {...routeProps} {...props} /> }/>
```
I think you also want to pass in props somewhere higher up, maybe in `BrowserRouter` ?

LeoForconesi: @rushtong, please check if my approach makes sense to you. This way it will pass props to each component that is present in Switch.

rushtong (approved):
Looks reasonable, thanks for the changes 👍

dags (approved):

PR #208, opened 2019-07-08
BTRX - 722 - Replace TinyMce

LeoForconesi:
## Addresses
[BTRX-722](https://broadinstitute.atlassian.net/browse/BTRX-722): Replace Tinymcs editor with a react implementation
[BTRX-502](https://broadinstitute.atlassian.net/browse/BTRX-502): Implement Comments tab on Project view. 
[BTRX-507](https://broadinstitute.atlassian.net/browse/BTRX-507): Implement Comments tab on Consent Group view. 

## Changes
- Use React library for tinyMce and remove old grails implementations.
- Make `Comments Controller` available as a resource. Replace Gorm call to get all comments to a sql query.
- Re-implement grails "comments" tab to react. On this code we introduced [React Bootstrap Table 2](https://react-bootstrap-table.github.io/react-bootstrap-table2/) which is the first step to update previous  version of this library

## Testing
- From an existing Project or Sample Data Cohort, navigate to `Comments` tab. Text area editor should be properly loaded, as the table of previous comments (if there is any) below this text area. 
- Enter any comment on the textarea and use styling from the toolkit, press Add button, the expected behaviour here is that the message appears in the table. The page shouldn't re load the entire browser window.
- If any error occurs while trying to add a new comment, an alert text should be visible indicating the user to try again later.
- Add button is disabled if the comment's textarea is empty.

---


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1383,10 +1386,19 @@ class QueryService implements Status {
         final String query =
                 ' select * from comment where project_key = :issueId'
         final SQLQuery sqlQuery = session.createSQLQuery(query)
-        final results = sqlQuery.with {
-            addEntity(Comment)
-            setString('issueId', issueId)
-            list()
+        Collection<Comment> results = null
```
rushtong: Drive by comment ... prefer to return an empty collection rather than null. Using null as a return value is generally not a great idea. When I see it - I try to clean it up.

grails-app/services/org/broadinstitute/orsp/CommentsService.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+            notifyService.sendComment(
+                    new NotifyArguments(
+                            toAddresses: toAddresses,
+                            fromAddress: fromAddress,
+                            subject: "Comment Entered: " + issue.projectKey,
+                            comment: comment.description,
+                            user: getUser(),
+                            issue: issue))
+            return comment
```
rushtong: You don't need an explicit return in groovy

rushtong: ... which I see is a copy/paste from the original, so just a cleanup task.

grails-app/services/org/broadinstitute/orsp/CommentsService.groovy
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     *
+     * @param issueId Project or Data Sample Cohort id
+     * @return  Collection of Comments associated to the given issueId
+     */
+    Collection<Comment> getCommentsForIssueId(String issueId) {
+        if (issueId) {
+            try {
+                Collection<Comment> comments = queryService.getCommentsByIssueId(issueId)
+                return comments
```
rushtong: same thing here - no need for a return.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: Got it!, will keep an eye on it if I see any i'll correct them too.

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -10,6 +10,9 @@ class UrlMappings {
         }
         // TODO: Would prefer a better way to redirect a root url to a controller regardless of action.
         // Without the extra declarations, the default index action is always the one called even if /funding is hit.
+
+        '/api/addComment'(controller: 'comments', action: 'saveNewComment', method: 'POST')
+        '/api/commentsList'(controller: 'comments', action: 'getComments', method: 'GET')
```
rushtong: The comments above this line actually refer to line 17 in this file.

LeoForconesi: Fixed.

LeoForconesi:
@rushtong @dags Please let me know if you consider that any other change is needed to be done here.

grails-app/controllers/org/broadinstitute/orsp/CommentsController.groovy
```diff
@@ -1,13 +1,57 @@
...
+            render([error: e.message] as JSON)
+        } catch (Exception e) {
+            response.status = 500
+            render([error: e.message] as JSON)
+        }
+    }
+
+    def getComments() {
+        JSON.registerObjectMarshaller(Comment) {
+            def output = [:]
```
rushtong: Two minor things here. 
1. This is actually a `LinkedHashMap<Object, Object>`, avoid def whenever possible. 
2. I see this pattern in UserController too, which is handy - but we're spreading out these marshallers and there's the potential for DRY violations this way. I'm wondering if these should be pulled up into a utility class (i.e. `UtilityClass.marshallObjects()`, `UtilityClass.marshallComment()` or something similar) so they can be more easily referenced from anywhere.

grails-app/controllers/org/broadinstitute/orsp/CommentsController.groovy
```diff
@@ -1,13 +1,57 @@
...
+        try {
+            Collection<Comment> comments = commentsService.getCommentsForIssueId(params.id)
+            response.status = 200
+            render comments as JSON
+        } catch(IllegalArgumentException e) {
+            response.status = 400
+            render([error: e.message] as JSON)
+        } catch(Error e) {
+            response.status = 500
+            render([error: e.message] as JSON)
```
rushtong: Not for this PR ... definitely a tech-debt ticket. We should abstract these error conditions even more than we're doing now. I think it would be nice to let the super-class handle the different common exception cases [the way we do in consent](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/Resource.java#L59). The object we render should also be an `Error` class that has a status and whatever exception is thrown.

grails-app/services/org/broadinstitute/orsp/CommentsService.groovy
```diff
@@ -0,0 +1,84 @@
...
+            try {
+                Collection<Comment> comments = queryService.getCommentsByIssueId(issueId)
+                comments
+            } catch (Exception e) {
+                log.error("An error has occurred when trying to get comments for issueId: ${issueId}.", e)
+                throw new Error()
+            }
+        } else {
+            log.error("Unable to get Comments from null issueId.")
+            throw new IllegalArgumentException()
```
rushtong: nit: Should have something in the exception, something like `Issue id is required.`

grails-app/services/org/broadinstitute/orsp/CommentsService.groovy
```diff
@@ -0,0 +1,84 @@
...
+     * @return  Collection of Comments associated to the given issueId
+     */
+    Collection<Comment> getCommentsForIssueId(String issueId) {
+        if (issueId) {
+            try {
+                Collection<Comment> comments = queryService.getCommentsByIssueId(issueId)
+                comments
+            } catch (Exception e) {
+                log.error("An error has occurred when trying to get comments for issueId: ${issueId}.", e)
+                throw new Error()
```
rushtong: nit: You're changing the exception here and the actual exception never gets passed up to the caller, so the error message going to the user is always empty. I suggest getting rid of the try/catch altogether and let the controller deal with what it gets back from the service class.

package.json
```diff
@@ -39,25 +39,30 @@
     "webpack": "2.4.1"
   },
   "dependencies": {
+    "@tinymce/tinymce-react": "^3.2.0",
```
rushtong: nit: I suggest pegging the version exactly - we've had problems (especially in consent-ui) with the compatible version annotations that pulled in updates that were actually incompatible. In other words, I don't trust NPM versioning promises - they have caused huge headaches in the past.

package.json
```diff
@@ -39,25 +39,30 @@
...
     "axios": "0.19.0",
     "babel-plugin-transform-class-properties": "6.24.1",
     "babel-polyfill": "6.26.0",
     "babel-register": "6.26.0",
     "bootstrap": "4.3.1",
+    "date-fns": "1.30.1",
+    "lodash": "4.17.11",
     "lodash-fp": "0.10.4",
+    "pdfmake": "^0.1.57",
```
rushtong: nit: peg version.

package.json
```diff
@@ -39,25 +39,30 @@
...
     "prop-types": "15.6.2",
     "react": "16.6.3",
     "react-bootstrap": "0.32.4",
     "react-bootstrap-table": "4.3.1",
+    "react-bootstrap-table-next": "^3.1.5",
+    "react-bootstrap-table2-paginator": "^2.0.7",
+    "react-bootstrap-table2-toolkit": "^2.0.1",
```
rushtong: mots ... (more of the same)

rushtong (approved):
Looks pretty good. Small changes recommended inline. Don't need to re-review if you want to tackle those.

grails-app/controllers/org/broadinstitute/orsp/CommentsController.groovy
LeoForconesi: 1- Def value is now refactored.
2- This makes sense, as you suggested I moved marshallers to a Utility class.

grails-app/controllers/org/broadinstitute/orsp/CommentsController.groovy
LeoForconesi: Indeed this would be an improvement. I created a tech debt story for this.
[BTRX-737](https://broadinstitute.atlassian.net/browse/BTRX-737)

grails-app/services/org/broadinstitute/orsp/CommentsService.groovy
LeoForconesi: This is now being handled as you suggested.

rushtong:
Still 👍

PR #209, opened 2019-07-10
Show Data Use Letter form to not authenticated users

vvicario:
## Addresses
There is not Jira ticket
## Changes
Include a high level summary of the changes and what they are intended for.

## Testing
To test this:
As a logged user, go to a Consent Group
In Documents tab, click on Add Document
In the Modal, select Data Use Letter type
Click on "Get shareable link"
Paste the link in an incognito window, or on other browser
Expected Result:
You should see the DUL form and be able to complete and submit it without being logged in the application.


grails-app/views/layouts/main_component.gsp
```diff
@@ -194,15 +194,23 @@
...
-    <div class="container">
-        <div id="login_spinner" class="hidden">
-            <div class="alert alert-success alert-dismissable" style="display: block">
-                Loading ... <span class="glyphicon glyphicon-refresh glyphicon-refresh-animate"></span>
-            </div>
-        </div>
-        <h3>About the ORSP Portal</h3>
-        <g:render template="/index/aboutBlurb"/>
-    </div>
+  <g:if test="${pageProperty(name:'body.dul') == 'true'}">
```
rushtong: This implementation is an anti-pattern. If we generalize this approach, then we have custom if blocks that are dependent on strings in unrelated gsps spread throughout the application - that is a brittle and un-refactor-able approach. There should be as close to zero custom logic in `main_component.gsp` as possible - `dataUseLetter/index.gsp` should be doing all of the work necessary for determining what kind of layout it is enclosed within. Instead, you should have a customizable `<meta name="layout" content="main_component"/>` in `dataUseLetter/index.gsp` that is smart enough to choose a parent layout that does not include the unwanted content.

vvicario:
@rushtong I created a new layaout for pages that needs to be open, please let me know if you have comments. Thanks

wloforteBroad:
@rushtong To test this:
1. As a logged user, go to a Consent Group
2. In Documents tab, click on Add Document
3. In the Modal, select Data Use Letter type
4. Click on "Get shareable link"
5. Paste the link in an incognito window, or on other browser

Expected Result:
You should see the DUL form and be able to complete and submit it without being logged in the application.

rushtong:
@wloforteBroad - The "Get Sharable Link" feature doesn't work for me. It says the link is copied to the clipboard, but nothing is copied to the clipboard. I can look at dev tools and see a dulToken, so something in the code must be using that to construct some kind of link.

Unrelated to this PR, the testing instructions bring up something that is strange to me. To get a shareable link, I have to click add document, even if there is already a DUL document on the consent group. That doesn't make any sense - the link sharing feature should be visible somewhere in the documents tab or on the DUL page. It could be visible only if there is a DUL, or grayed out in the case that it isn't. I think that should be addressed separately - but please prioritize with Andrea.

rushtong:
@wloforteBroad - Tested the shareable link feature in chrome and it does work - just not in firefox. Will continue testing.

grails-app/views/layouts/unguarded_component.gsp
```diff
@@ -0,0 +1,119 @@
...
+        <g:render template="/index/aboutBlurb"/>
+    </div>
+</auth:nonBroadSession>
+
+<auth:broadSession>
+    <div class="container">
+        <g:render template="/base/messages" />
+        <g:layoutBody/>
+    </div>
+</auth:broadSession>
```
rushtong: The three auth blocks above can be condensed into one. Since the page doesn't show different content for different users, all three can just be:
```
<div class="container">
    <g:render template="/base/messages" />
    <g:layoutBody/>
</div>
```

grails-app/views/layouts/unguarded_component.gsp
```diff
@@ -0,0 +1,119 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+    <title><g:layoutTitle default="ORSP Portal"/></title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="ORSP Portal">
+    <meta name="author" content="Business Systems">
```
rushtong: Copy paste - can probably just remove the author line

grails-app/views/layouts/unguarded_component.gsp
```diff
@@ -0,0 +1,119 @@
...
+    }
+    @media (min-width: 765px) and (max-width: 1200px) {
+        body {
+            margin-top: 110px;
+        }
+    }
+    </style>
+    <script src="https://apis.google.com/js/platform.js" async defer></script>
+    <meta name="google-signin-client_id" content="${grailsApplication.config.googleSignInClientId}">
+    <script>
```
rushtong: This script is now in three places. It would be nice to refactor that into a common function so we're not increasing our tech-debt.

grails-app/views/layouts/unguarded_component.gsp
rushtong: One correction to the above ... the about page link might need to be updated based on recent conversations around when to show that.

wloforteBroad:
@rushtong We did review this in the past. We refine this and demo it as well. I would like to discuss this some more. Ping me if you have time.

vvicario:
@rushtong when you have some time, could you review it? Thanks!

rushtong:
@vvicario, @wloforteBroad Thank you for the changes - the code, in general, looks good. There is absolutely a security bug, however, and I don't have enough time to dig through it to figure out exactly where and why it's happening. Both authed and unauthed requests are skipping `AuthenticatedController` when calls to  `DataUseLetterController -> show()` are made. Since `AuthenticatedController` implements `Interceptor`, its `before` method should be returning false, but it is never getting hit in either user-flow. It might be the case that this happens elsewhere too, I have not investigated that deeply, but it is clearly a bug in this workflow. The underlying bug needs to be identified and at least ticketed before moving forward with this approach. I wonder if it has something to do with the `@Resource` annotation.

@vvicario, Please update the PR description as the template asks - that makes it a lot easier to do a quick review.

vvicario:
@rushtong I updated ApiInterceptor to include all the controllers that we need about security, it's working, I didn't include dataUseLetter controller because it should be acceded without authentication. For testing purpose I put dataUseLetter controller but I got forbidden error wherever I am not authenticated. 
On the other hand, I tried  authenticated controller like this: " match controller: 'authenticated'", but it didn't work as expected. So, I have to put one by one.

grails-app/controllers/org/broadinstitute/orsp/ApiInterceptor.groovy
```diff
@@ -10,7 +10,25 @@ import grails.converters.JSON
...
+        match controller: 'sampleConsentLink'
+        match controller: 'consentGroup'
+        match controller: 'dataUse'
+        match controller: 'issueList'
+        match controller: 'history'
+        match controller: 'irb'
+        match controller: 'ne'
+        match controller: 'nhsr'
+        match controller: 'search'
+        match controller: 'status'
```
rushtong: I don't think this should be intercepted - it is an un-authed page that shows you this: http://orsp.broadinstitute.org/api/status

rushtong: And totally unrelated, the notification status looks like it has a bug
```
{
  "ok": false, 
  "systems": {
    "BSP": {
      "ok": true
    }, 
    "Crowd": {
      "ok": true
    }, 
    "DUOS": {
      "ok": true
    }, 
    "DataBioOntology": {
      "ok": true
    }, 
    "Database": {
      "ok": true
    }, 
    "GCS": {
      "ok": true
    }, 
    "Notifications": {
      "messages": [
        "No status available from Notifications Service"
      ], 
      "ok": false
    }, 
    "Ontology": {
      "ok": true
    }
  }
}
```

PR #211, opened 2019-07-16
 [BTRX-734-735-736] Data use Letter Link, PDF Box Update and Alert Time extension

wloforteBroad:
## Addresses
- https://broadinstitute.atlassian.net/browse/BTRX-734
- https://broadinstitute.atlassian.net/browse/BTRX-735
- https://broadinstitute.atlassian.net/browse/BTRX-736

## Changes
- Update PDF Box version to 2.0.15
- Sample/Data Cohort creation alert time was extended to 15 seconds as per ORSP team request.
- The Data Use Letter functionality to create the shareable link or send it by email was moved from the Documents modal to the Documents page in a new Panel.

## Testing
- Create a Sample/Data Cohort and the alert will appear.
- In a Consent Group, in the Documents page, the new Panel called Data Use Letter should appear and the user can perform the related actions from there.

---


src/main/webapp/components/Documents.js
```diff
@@ -32,14 +38,94 @@ export const Documents = hh(class Documents extends Component {
...
+      error: false,
+      alertType: 'success',
+      invalidEmail: false,
+      collaboratorEmail: '',
     }
     this.removeDocument = this.removeDocument.bind(this);
+    this.getShareableLink = this.getShareableLink.bind(this);
   }
 
+  getShareableLink = () => {
```
rushtong: Curious about why we're adding all of this functionality to the `Documents` component and really inflating that to handle a lot more than it really should. It seems to be this could be broken down into its own small component for easier encapsulation.

src/main/webapp/components/Documents.js
```diff
@@ -144,10 +243,57 @@ export const Documents = hh(class Documents extends Component {
         })
       ]),
 
+      Panel({ title: "Data Use Letter" }, [
```
rushtong: This is what I think should be broken out into a new component.

wloforteBroad: Sounds good, will do

PR #213, opened 2019-07-18
Only Render DUL Panel in CG

wloforteBroad:
## Addresses
No ticket

## Changes
Added isRender property to DUL panel to make it visible only in CG Documents

## Testing
DUL Panel should not appear in Project Documents page

---


wloforteBroad:
Thanks all!

PR #215, opened 2019-07-22
BTRX-728 - ORSP: Implement "Awaiting Action from" feature in the new flow

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-728

## Changes
Add actor issue extra property when specific actions are performed
## Testing
Follow new flow actions specified in this doc:
https://docs.google.com/spreadsheets/d/1J42uxJpkCxX8Pu2xFCR8T5-wshYJRpJyrXi_GfhSMX8/edit?usp=drive_web&ouid=110074181392966592330

---


grails-app/services/org/broadinstitute/orsp/TransitionService.groovy
```diff
@@ -96,16 +97,40 @@ class TransitionService {
      * NE/NHSR projects go into Submitting to ORSP
      */
     @Transactional
-    handleIntake(Issue issue, Collection<String> actors, String status, String author) throws DomainException {
-        def deletableProperties = [IssueExtraProperty.ACTOR]
-        deleteProps(issue, deletableProperties)
+    handleIntake(Issue issue, Collection<String> actors) throws DomainException {
```
rushtong: Just a note that this class is heavily documented for the express purpose of giving developers a clear overview of that statuses are required for different project types.

vvicario:
@rushtong TransitionService has a lot of code that is not valid any more, I think we have to create a story to remove old code after we finish with migration. 
We are not handling status changes in TransitionService. Methods like submit, modify, accept, etc are been used in old controllers (Nhsr, Ne, Irb) and those don't represent the new flow.

PR #218, opened 2019-07-24
BTRX-621 ORSP: Convert Profile Page

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-621
## Changes
Migrate Profile page

## Testing
Login and click on your name, check your personal information
---


src/main/webapp/profile/Profile.js
```diff
@@ -0,0 +1,90 @@
...
+  init = () => {
+    spinnerService.showAll();
+    User.getUserSession(component.getUserUrl).then(resp => {
+      let role = 'Standard Access';
+      if (resp.data.isAdmin || resp.data.isORSP || resp.data.isComplianceOffice) {
+        role = 'Admin';
+      } 
+      else if (resp.data.isViewer) {
+        role =  "Read Only Access";
+      }
```
rushtong: Instead of setting role as a state value, it would be more efficient to take this role logic and turn it into a function. Then you don't have to store it as a state value. Same thing with `lastLoginDate`. Its better to store as little as possible in state and rely on calculating values using functions that take a state value as input. In this case, all of this can be calculated from `state.user`.

PR #220, opened 2019-07-29
Fix Awaiting action

vvicario:
## Addresses
No Jira Ticket
## Changes
Removing actor when it's needed and refresh information in project status box
## Testing
After uploading files or approving project information, awaiting action should display the correct user
---


src/main/webapp/main/Main.js
```diff
@@ -44,9 +44,9 @@ class Main extends Component {
 
   updateDetailsStatus = (status) => {
     this.setState(prev => {
-      prev.status.projectReviewApproved = status.projectReviewApproved;
-      prev.status.summary = status.summary;
-      prev.status.actor = status.pm;
+      prev.status.projectReviewApproved = status.extraProperties.projectReviewApproved;
+      prev.status.summary = status.issue.summary;
+      prev.status.actor = status.extraProperties.actor;
```
rushtong: I think there is a solid chance of some of these values being null, especially for legacy projects.

vvicario: Greg, I added validations to check null cases

vvicario:
@rushtong Can you check it again? Thanks!

PR #222, opened 2019-07-31
[BTRX-731] Convert submission tab

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-731

## Changes
Replace submissions tab for react code.

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -21,6 +22,16 @@ class SubmissionController extends AuthenticatedController {
                 ])
     }
 
+    def getSubmissions() {
+        Issue issue = queryService.findByKey(params.id)
+        if (issueIsForbidden(issue)) {
+            redirect(controller: 'Index', action: 'index')
```
rushtong: We should never be redirecting inside a controller. The error should be returned and the UI needs to deal with it.

src/main/webapp/components/Table.js
```diff
@@ -128,6 +139,31 @@ export const Table = hh(class Table extends Component {
     }, ["Info Link"])
   };
 
+  redirectToDocumentLink = (cell, row) => {
```
rushtong: nit: the name first led me to think this was redirecting the user to a document download link but it's really just a `documentLink` ... there's not redirecting happening here.

src/main/webapp/components/Table.css
```diff
@@ -68,3 +68,11 @@
...
     cursor: pointer;
 }
+
+.submission-comment {
+    margin: 10px;
+}
+
+.submission-download {
+    margin-right: 10px;
+}
```
rushtong: nit: These could be in-lined style objects in `Table.js` instead of here.

src/main/webapp/projectContainer/Submissions.css
```diff
@@ -0,0 +1,27 @@
...
+    text-overflow: initial;
+    word-break: break-word;
+}
+
+.linkOverflowEllipsis {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    color: #337ab7;
+}
```
rushtong: nit: All of these should be in-lined style objects in `Submissions.js`.

src/main/webapp/projectContainer/Submissions.js
```diff
@@ -1,47 +1,153 @@
...
+    const indexButton = a({
+      className: 'btn btn-default btn-xs pull-left link-btn',
+      href: `${component.contextPath}/submission/index?projectKey=${component.projectKey}&submissionId=${data.id}`
+    }, [component.isAdmin === true ? 'Edit': 'View']);
+    const submissionComment = span({className: 'submission-comment'}, [data.comments]);
+    return h(Fragment, {}, [indexButton, submissionComment]);
+  };
+
+  redirectNewSubmission(e) {
+    window.location.href=`${component.serverURL}/api/submissions/add-new?projectKey=${component.projectKey}&type=${e.target.id}`;
```
rushtong: Do we really have to redirect here? We're trying very hard to navigate to places instead of redirect. This pattern breaks the SPA model.

grails-app/controllers/org/broadinstitute/orsp/api/FileHelperController.groovy
```diff
@@ -12,6 +12,8 @@ import org.broadinstitute.orsp.StorageDocument
 import org.broadinstitute.orsp.SupplementalRole
 import org.springframework.web.multipart.MultipartFile
 
+import javax.swing.text.Document
```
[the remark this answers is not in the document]
rushtong: Nice catch. Interesting that this even shows up as a valid dependency - we're not doing anything swing-related here.

src/main/webapp/projectContainer/Submissions.js
```diff
@@ -1,47 +1,151 @@
...
 
-  getSubmissions() {
-    ProjectMigration.getSubmissions(component.projectKey).then(resp => {
+  getDocumentLink = (data) => {
+    return [component.serverURL, 'api/files-helper/get-document?id=' + data].join("/");
+  };
+
+  submissionEdit = (data) => {
+    const indexButton = a({
+      className: 'btn btn-default btn-xs pull-left link-btn',
```
rushtong: It would be nice if these were also inline styles

src/main/webapp/projectContainer/Submissions.js
```diff
@@ -1,47 +1,151 @@
...
-      buttons: [],
-      language: { search: 'Filter:' },
-      pagingType: "full_numbers",
-      pageLength: 50,
-      columnDefs: [{ targets: [1, 2], orderable: false }],
-      order: [0, "desc"]
-    });
-    $("#submission-tabs").tabs();
+  redirectNewSubmission(e) {
+    window.location.href = `${component.serverURL}/api/submissions/add-new?projectKey=${component.projectKey}&type=${e.target.id}`;
```
rushtong: Do we have to redirect here? Can we not load them? Any kind of location change like this breaks the current context.

src/main/webapp/projectContainer/Submissions.js
```diff
@@ -1,47 +1,151 @@
...
-      div({dangerouslySetInnerHTML: { __html: this.state.content } },[])
+      div({}, [
+        button({
+          className: "btn buttonPrimary floatRight",
+          style: { 'marginTop': '15px' },
+          onClick: '',
+          isRendered: this.state.readOnly === true && !component.isViewer
+        }, ["Edit Information"]),
+        Panel({title: "Submissions"}, [
+          div({ className: "containerBox" }, [
```
rushtong: More class names here that could probably be inline styles

src/main/webapp/projectContainer/Submissions.js
tadeork: I think this will be solved when I complete https://broadinstitute.atlassian.net/browse/BTRX-742

src/main/webapp/projectContainer/Submissions.js
tadeork: I think in this case it makes no sense to put Bootstrap styles inline in the component.

rushtong: 👍 since these are bootstrap styles. Relatedly, one of the goals is to move away from shared css altogether - it is very difficult to maintain and introduces a host of style bugs over time.

src/main/webapp/projectContainer/Submissions.js
rushtong: Sounds good 👍

src/main/webapp/projectContainer/Submissions.js
tadeork: Yes, I agree on that.

PR #224, opened 2019-08-01
[BTRX-539] Replace constants

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-539

## Changes
Replaced URL strings to constant strings and changed methods in the components to meet the new approach. There are many files changed because of this reason.

## Testing

---


src/main/webapp/util/UrlConstants.js
```diff
@@ -0,0 +1,71 @@
+const context = process.env.NODE_ENV !== 'test' ? component.contextPath : 'http://localhost';
```
rushtong: nit: I generally prefer non-negative logic blocks when possible - it's just as easy and more legible to invert this:
```suggestion
const context = process.env.NODE_ENV === 'test' ? 'http://localhost' : component.contextPath;
```

PR #225, opened 2019-08-02
ORSP: Implement mock-ups for Data Use Letter Link

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-741

## Changes
Update Consent Group -> Documents tab. Including Sample Collection date range and repository deposition question. Populate DUL form with the information just added in documents tab
## Testing
Go to:
1) Consent Group -> Documents tab. Fill in  Data Use Limitation Record Request information and get shareable link.
2) Go to the link. Information that you added in Data Use LImitation Record must be populated

---


grails-app/views/mainContainer/index.gsp
```diff
@@ -1,8 +1,15 @@
 <html>
   <head>
+
+  <g:if test ="${openUrl}">
+    <meta name="layout" content="unguarded_component"/>
+  </g:if>
```
rushtong: I don't like conditional layout rendering this way. Conditional content, yes, but layouts are more brittle this way. This index page should be agnostic about *how* it is invoked. If there needs to be a different kind of layout, then it should be in a different gsp. Better to have a separate gsp that uses this layout.

src/main/webapp/components/Documents.css
```diff
@@ -1,34 +1,42 @@
...
+.alertMessage.alert-success {
+    margin: 20px 0 10px 0;
+}
+
+.getShareable {
+    margin-top: 20px;
+}
+
+.getShareable .glyphicon {
+    margin-right: 5px;
```
rushtong: It would be a lot nicer if we used inline js style objects. Shared css like this is super brittle. I know we've gone down this path for a while, but if possible, we should move in the inline direction.

vvicario:
@rushtong Can you ckeck it please? Thanks!

PR #226, opened 2019-08-05
BTRX - 733 - Convert Sample Data Cohorts Tab

LeoForconesi:
## Addresses
[BTRX-733](https://broadinstitute.atlassian.net/browse/BTRX-733): Convert Sample/Data Cohort Tab
## Changes
- Remove deprecated grails view
- Used [rc-collapse](https://github.com/react-component/collapse) library for collapse functionallity. 
- Created a generic collapsible component where we can put custom components as header and body
- Refactored consent groups list list action
- First use of react redirections. This threw a memory leak warning for trying to update state for unmounted components caused by spinner-service. This is fixed by unsubscribing a specific spinner created by the component using it. 
- Corrected table buttons text to match as they were in old implementations.
**Note:** We still need to check all the places where spinnerService is used, I only fixed the places where ProjectReview / Sample Data Cohorts / New Consent Group / Existing Consent group redirections are, I'm pretty sure there are more, but escapes to this PR scope and size so it would be nice to catch in a sepparate story.
**Note2:**We are still depending on grails redirections somehow, a lot of [these](https://github.com/broadinstitute/orsp-pub/compare/BTRX-733-convertSampleDataCohortsTab#diff-8ed4190862442832525e51e9360e0ab4R42) main_layout redirections will disapear as we advance in react migration.
## Testing
- Create a new project and navigate to Sample/Data Cohorts tab.
- Click on `Add New Sample/ Data Cohorts` button. (this redirection is made by react),
- Refresh page using F5 or browser's button. This shouldn't throw any warning in console.
- Click on browser's back button (return to previous page), this shouldn't throw any warning in console. (Note: Consent group has another async setState not beign handled here and not related to spinnerService)
- Return to `Add New Sample/ Data Cohorts` and create a new one, be sure to add some documents in order to later check them in Project's Review page
- After previous creation, a green message should appear indicating its creation
- The new element should appear in the tab. Check their buttons, Approve, Reject, Unlink, Request Clarification.
- Press the drop down button, the documents that you uploaded should appear here. The name of the file should be clickable, downloading the file.
- Check that Delete button is working
- Only Admin users can Approve, Unlink, Reject and Delete attachments
- Since Spinner Service has been modified, the places to be checked if the spinner still working are:
  . Access Project Review, and Consent group (loading indicator appears)
  . Project Review Admin Only Tab (update some date here and save)
  . Project Review, and Consent group Comments Tab ( add a new comment)
  . Project Review, and Consent group Documents Tab ( upload a new file)
---


grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -41,17 +41,23 @@ class UrlMappings {
         '/api/consent-group/sample-collections'(controller: 'consentGroup', action: 'getConsentGroupSampleCollections')
         '/consent-group/name-search'(controller: 'consentGroup', action: 'getConsentGroups')
         '/consent-group/sample-search'(controller: 'consentGroup', action: 'unConsentedSampleCollections')
+
```
tadeork: Remove spacing.

grails-app/controllers/org/broadinstitute/orsp/api/FileHelperController.groovy
```diff
@@ -130,4 +130,13 @@ class FileHelperController extends AuthenticatedController{
             }
         }
     }
+
+    def deleteDocumentByUuid() {
+        if (params.uuid) {
```
tadeork, line 135: Catch null uuid.

vvicario (commented):
Please verify all new lines

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -41,17 +41,23 @@ class UrlMappings {
...
+        '/consent-group/use-existing'(controller:'newConsentGroup', action:'redirect')
+        '/consent-group/new'(controller:'newConsentGroup', action:'redirect')
+
         '/api/consent-group/approve'(controller: 'newConsentGroup', action: 'approveConsentGroup')
         '/api/consent-group'(controller: 'newConsentGroup', action: 'update', method: 'PUT')
         '/api/consent-group/review'(controller: 'consentGroup', action: 'getConsentGroup', method: 'GET')
         '/api/consent-group/upload-modal'(controller: 'consentGroup', action: 'loadModalWindow', method: 'GET')
         '/api/consent-groups'(controller: 'consentGroup', action: 'projectConsentGroups', method: 'GET')
+        '/api/consent-group/getProjectConsentGroups'(controller:'newConsentGroup', action:'getProjectConsentGroups', method: 'GET')
```
vvicario: Please use the same url pattern

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1278,6 +1278,18 @@ class QueryService implements Status {
         results
     }
 
+     StorageDocument findAttachmentByUuid(String uuid) {
+        final session = sessionFactory.currentSession
+        final String query = ' select * from storage_document where uuid = :uniqueId '
+        final sqlQuery = session.createSQLQuery(query)
+        StorageDocument result = sqlQuery.with {
```
vvicario:
```suggestion
        sqlQuery.with {
            addEntity(StorageDocument)
            setParameter("uniqueId", uuid)
            uniqueResult()
        }
```

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -78,6 +80,27 @@ class UtilityClass {
...
+                return [
+                        id             : issue.id,
+                        projectKey     : issue.projectKey,
+                        type           : issue.type,
+                        status         : issue.status,
+                        summary        : issue.summary,
+                        reporter       : issue.reporter,
+                        requestDate    : issue.requestDate,
+                        attachments    : issue.attachments,
+                        title          : issue.summary,
```
vvicario: remove ","

src/main/webapp/util/TableUtil.js
```diff
@@ -61,3 +62,18 @@ function parseDataElements(el, key) {
...
+export const formatUrlDocument = (file) => {
+  return {
+    href: `${UrlConstants.downloadDocumentUrl}?uuid=${file.uuid}`,
+    target: '_blank',
+    title: file.fileName
+  }
+};
+
+export const parseDate = (date) => {
+  if (date !== null) {
```
vvicario: we can use isEmpty to validate date

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -496,4 +497,20 @@ class ConsentService implements Status {
...
+            Issue issue = queryService.findByKey(projectKey)
+            Collection<ConsentCollectionLink> collectionLinks = ConsentCollectionLink.findAllByProjectKey(issue.projectKey)
+            Map<String, ConsentCollectionLink> collectionLinksMap = collectionLinks?.collectEntries{[it.consentKey, it]}
+            Collection consentGroups = queryService.findByKeys(collectionLinksMap)
+            result = [
+                    issue        : issue,
+                    consentGroups: consentGroups?.sort { a, b -> a.summary?.toLowerCase() <=> b.summary?.toLowerCase() }
+            ]
+        } else {
+            throwConsentException("Error trying to get Project's Consent groups: Empty projectKey")
```
vvicario: I think you can throw IllegalArgumentException instead of ConsentException

src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
```diff
@@ -0,0 +1,91 @@
...
+              span({ className: 'req-tooltip' },[ 'Request Clarification']),
+              span({ className: 'arrow-down' },[])
+            ]),
+            this.linkToConsentGroup(projectKey, summary)
+          ])
+        ])
+      ])
+    )
+  }
+});
```
tadeork: EOF line.

src/main/webapp/components/TableComponent.js
```diff
@@ -51,15 +51,19 @@ export const TableComponent = hh(class TableComponent extends Component {
...
                 remote= {{
                   filter: remoteProp,
                   pagination: remoteProp,
                   sort: remoteProp,
                   cellEdit: false
                 }}
-                pagination={ paginationFactory({ page: this.props.page, totalSize: this.props.totalSize, sizePerPageList: this.props.sizePerPageList  })}
+                pagination= {
+                  this.props.pagination ?
+                  paginationFactory({ page: this.props.page, totalSize: this.props.totalSize, sizePerPageList: this.props.sizePerPageList  })
```
tadeork: Use multi lines for attributes. It's more legible.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -508,15 +508,17 @@ class NewConsentGroup extends Component {
         showSubmit: this.showSubmit,
         submitHandler: this.submitNewConsentGroup,
         disabledSubmit: this.state.formSubmitted,
+
       }, [
           NewConsentGroupGeneralData({
             title: "Sample/Data Cohort Info",
             currentStep: currentStep,
             user: this.state.user,
+
```
tadeork: Spacing.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -508,15 +508,17 @@ class NewConsentGroup extends Component {
         showSubmit: this.showSubmit,
         submitHandler: this.submitNewConsentGroup,
         disabledSubmit: this.state.formSubmitted,
+
```
tadeork: Spacing.

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -64,11 +159,10 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
     this.closeRequestClarification();
   };
 
-  loadConsentGroups = (rThis) => {
-    $('.consent-group-panel-body').hide();
+  collapseBtnAnimationListener() {
     $('.consent-accordion-toggle').on('click', function () {
-      var icon = $(this).children().first();
-      var body = $(this).parent().parent().next();
+      let icon = $(this).children().first();
```
tadeork: I think this could be 
```suggestion
      const icon = $(this).children().first();
```

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -64,11 +159,10 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
...
+      let body = $(this).parent().parent().next();
```
tadeork: Same here.      
```suggestion
      const body = $(this).parent().parent().next();
```

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -77,82 +171,140 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
...
+            requestClarificationHandler: this.requestClarification
+          }
+        }
       });
     }
+    return parsedArray;
+  };
+
+  redirect = (action) => {
+    const path = action === 'new' ? '/consent-group/new?projectKey='+ component.projectKey : '/consent-group/use-existing';
```
tadeork:
```suggestion
    const path = action === 'new' ? UrlConstants.updateConsentGroupUrl +?projectKey='+ component.projectKey : UrlConstants.consentGroupExistingUrl;
```
This last one should be added to UrlConstants

tadeork (commented):
Please see comments in line.

src/main/webapp/components/Documents.js
```diff
@@ -215,11 +223,9 @@ export const Documents = hh(class Documents extends Component {
...
       ]),
-
       h(Spinner, {
-        name: "mainSpinner", group: "orsp", loadingImage: component.loadingImage
+        name: DOCUMENTS_SPINNER, group: "orsp", loadingImage: component.loadingImage
       })
-
     ])
   }
 });
```
tadeork: EOF line.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1278,6 +1278,18 @@ class QueryService implements Status {
         results
     }
 
+    @SuppressWarnings("GroovyAssignabilityCheck")
+    StorageDocument findAttachmentByUuid(String uuid) {
+        final session = sessionFactory.currentSession
+        final String query = ' select * from storage_document where uuid = :uniqueId '
```
rushtong: This will also find deleted documents

rushtong: `uniqueId` is misleading, that could be anything. Suggest using the actual thing passed in:
```suggestion
        final String query = ' select * from storage_document where uuid = :uuid '
```

src/main/webapp/util/Utils.js
```diff
@@ -49,6 +49,10 @@ export const handleRedirectToProject = (serverURL, projectKey) => {
   return [serverURL, "project", "main?projectKey=" + projectKey + "&tab=review"].join("/");
 }
 
+export const handleRedirectToConsentGroup = (serverURL, consentKey, projectKey) => {
```
rushtong: Is this really a redirect or is it just a url?

rushtong: Or, to be more clear, this isn't a handler, it's just a url.

src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
```diff
@@ -0,0 +1,91 @@
+import React, { Component } from 'react';
+import { button, hh, span, div, i, a } from 'react-hyperscript-helpers';
+import { handleRedirectToConsentGroup, isEmpty } from '../util/Utils';
+import '../components/Btn.css';
+
+const POINTER = {
```
rushtong: nit, naming. Would read better if this were a styles object with pointer being one of the objects in it.
```suggestion
const styles = {
   pointer: { ... }
```

src/main/webapp/components/Btn.css
```diff
@@ -67,6 +67,20 @@
...
+.rc-collapse .rc-collapse-item.collapsible .rc-collapse-header,
+.rc-collapse .rc-collapse-item.collapsible .rc-collapse-header:focus {
+  padding-bottom: 0;
+  outline: none;
+}
+
+.rc-collapse {
+  margin-bottom: 6px;
+}
+
```
rushtong, line 79: Would be nicer if these were all inline styles

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -123,6 +125,10 @@ export const ConsentGroupReview = hh(class ConsentGroupReview extends Component
     this.init();
   }
 
+  componentWillUnmount() {
+    spinnerService._unregister(CONSENT_GROUP_REVIEW_SPINNER);
+  }
```
rushtong, line 130: I'm not liking this pattern of requiring that every component un-register itself. It's boilerplate that any new components will need to copy/paste. Would be much better if we could use a HOC for this and wrap the components with a spinner HOC.

rushtong (changes requested):
Minor comments inline, except for the spinner service. Every component that needs a spinner should have *its own spinner* instead of using a global spinner. That way, each one controls only its own version of the spinner and no need to track the different states of different spinners by name. The current pattern is pretty brittle and subject to bugs when new components are created, old ones refactored, etc.

See [React HOCs](https://reactjs.org/docs/higher-order-components.html) and [HOCs by example](https://levelup.gitconnected.com/understanding-react-higher-order-components-by-example-95e8c47c8006) for some guidance on how to do that.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
LeoForconesi: That's true. I've modified this query like this ->
 
`SELECT * FROM storage_document WHERE uuid = :uuid AND deleted = 0`

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -39,6 +38,10 @@ class NewConsentGroupController extends AuthenticatedController {
         render(view: "/mainContainer/index", model: [projectKey: params.projectKey, consentKey: params.consentKey, issueType: 'consent-group'])
     }
 
+    def redirect() {
```
rushtong: nit: confusing name. This doesn't actually redirect, it just renders a view.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
```diff
@@ -168,6 +171,24 @@ class NewConsentGroupController extends AuthenticatedController {
...
+            JSON.use(UtilityClass.ISSUE_COMPLETE) {
+                render( consentGroups as JSON)
+            }
+        } catch(IllegalArgumentException e) {
+            response.status = 400
+            render([message: 'Error while trying to get project\'s Consent Groups. ' + e as JSON])
+        } catch(Exception e) {
+            log.error("Error while trying to get Consent Groups for ProjectKey: ${params.projectKey}. ", e.getMessage())
+            response.status = 500
+            render([message: 'Error while trying to get project\'s Consent Groups.' as JSON])
```
rushtong: nit here and above, you can avoid the escape, `\`, if you double quote the string.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -78,6 +79,26 @@ class UtilityClass {
         }
     }
 
+    /**
+     * Register Issue's JSON mapping to return its transient properties
+    */
+    static void registerIssueForSampleDataCohortsMarshaller() {
```
rushtong: nit: This is mostly an expansion of `registerIssueMarshaller` minus the `reviewCategory` stuff. Why not just add the new things to `registerIssueMarshaller` and use that? Having two different DTOs that are essentially the same thing is a maintenance headache. You'll have two different APIs that essentially do the exact same thing. I strongly recommend simplification and re-use instead of duplication.

rushtong:
I know you're working on the spinner piece, but it might be better to tackle that as a separate story (I leave that decision to the team). If you do, this article describes an implementation approach that is less brittle and more flexible than the current approach of storing `N` spinners in global state. 

https://medium.com/@Farzad_YZ/handle-loadings-in-react-by-using-higher-order-components-2ee8de9c3deb

LeoForconesi:
Regarding spinner subscribe and unsubscribe pattern. 
I've created a HOC which will handle a unique spinner in the entire app through spinnerService `showMain()` and `hideMain().`

The only problem here is that only components accesed through React's router can reach this spinner. Eventually we will have everything under this Hoc, but meanwhile there are some components (like NewProject) that are accesed directly from a grails view, giving as result that the main spinner is never created. 

I was gonna propose create a new story, so we are agree on that. I left the [base](https://github.com/broadinstitute/orsp-pub/pull/226/files#diff-41cc7ec30f3c5dcb1a6a8656e512b919R10) to do it as you suggested in your last comment.

grails-app/controllers/org/broadinstitute/orsp/api/NewConsentGroupController.groovy
LeoForconesi: I've changed this to `renderMainComponent`.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
LeoForconesi: I agree. I've removed this unnecessary marshaller,

rushtong:
@LeoForconesi 
> I've created a HOC which will handle a unique spinner in the entire app through spinnerService showMain() and hideMain().

The showMain and hideMain functions are not HOCs. Have a read through the links posted. 

Feel free to ping me when you think this is ready for re-review.

LeoForconesi:
As we previously agreed, I did a revert on the code on the spinner handling, this will be attended in [BTRX-754](https://broadinstitute.atlassian.net/browse/BTRX-754) : ORSP - Fetch Spinner to a HOC

rushtong (approved):
👍 With the caveat that we address the spinner issue in a follow-on PR.

vvicario (approved):

PR #227, opened 2019-08-06
Btrx 600 Rewrite User Sync 

dags:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-600

## Changes
Service was refactored to read users from BigQuery user database instead of internal Crowd service. 

## Testing
Hitting app_url + 'user/sync' should read users from BigQuery and add new users to ORSP user table, if any.

---


grails-app/services/org/broadinstitute/orsp/CrowdService.groovy
```diff
@@ -1,131 +1,113 @@
...
-                log.error("Content-Type: ${resp.headers.'Content-Type'}")
+        // Instantiate a client.
+        BigQuery bigquery =
+                BigQueryOptions.newBuilder()
+                        .setCredentials(getCredential())
+                        .build()
+                        .getService()
+
+        QueryJobConfiguration queryConfig = QueryJobConfiguration
+                .newBuilder("SELECT * FROM `broad-bits.data_warehouse.people`")
```
rushtong: See https://cloud.google.com/bigquery/docs/best-practices-costs
We're only looking for `broad_email`, `username`, `first_name`, and `last_name` ... we should only query that data.
```suggestion
                .newBuilder("SELECT broad_email, username, first_name, last_name FROM `broad-bits.data_warehouse.people`")
```

grails-app/services/org/broadinstitute/orsp/CrowdService.groovy
```diff
@@ -1,131 +1,113 @@
...
+        } else if (queryJob.getStatus().getError() != null) {
+            log.error(queryJob.getStatus().getError().toString())
+        } else {
+            // Get the results.
+            TableResult result = queryJob.getQueryResults()
+            // iterate over results to find out missing users
+            for (FieldValueList row : result.iterateAll()) {
+                String email = row.get("broad_email").getStringValue()
+                String userName = row.get("username").getStringValue()
+                if (email && !filterUsers.contains(userName)) {
```
rushtong: To make this class more useful, I would break out these two concepts out into two different functions. By mixing Find-All and Filter in one function, we lose the ability to do more interesting things with the data.

1. Find all Broad Users
2. Filter user list against list of known users

grails-app/services/org/broadinstitute/orsp/CrowdService.groovy
dags: Ok, thanks for the link !

src/test/groovy/org/broadinstitute/orsp/config/CrowdConfigurationSpec.groovy
```diff
@@ -15,11 +15,7 @@ class CrowdConfigurationSpec extends Specification implements ServiceUnitTest<Cr
         CrowdConfiguration config = service.crowdConfiguration
 
         expect:
-        !config.url.isEmpty()
-        !config.username.isEmpty()
-        !config.password.isEmpty()
-        config.interval > 0
-        config.range > 0
+        !config.config.isEmpty()
```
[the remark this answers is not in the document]
dags: done

grails-app/services/org/broadinstitute/orsp/CrowdService.groovy
dags: makes sense.

PR #231, opened 2019-08-07
ORSP: Convert Data Use Restrictions Tables

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-747

## Changes
Migrate Data Use Restriction Report
## Testing
Go to Admin ->  Data Use Restriction Report
---


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1486,4 +1486,87 @@ class QueryService implements Status {
...
+        if (pagination.searchValue) {
+            query = query + 'where consent_group_key LIKE :term OR vault_export_date LIKE :term '
+        }
+        SQLQuery sqlQuery = getSessionFactory().currentSession.createSQLQuery(query)
+        if (pagination.searchValue) sqlQuery.setString('term', pagination.getLikeTerm())
+        Long totalRows = sqlQuery.uniqueResult()
+        if (orderColumn) {
+            query = query + " order by " + orderColumn + pagination.sortDirection
+        }
+        sqlQuery = getSessionFactory().currentSession.createSQLQuery(query.replace('count(id)', '*'))
```
rushtong: Create new instead of mutating the existing SQLQuery

rushtong: The code would read a lot cleaner and less susceptible to future bugs if you had two separate query strings. I would suggest first creating a count query string, then creating a star query string from that and using the two separate query strings in two separate `SQLQuery`s.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1486,4 +1486,87 @@ class QueryService implements Status {
...
+
+    PaginatedResponse findDataUseRestrictions(PaginationParams pagination) {
+        String query = 'select count(id) from data_use_restriction '
+        String orderColumn = getRestrictionOrderColumn(pagination.orderColumn)
+        if (pagination.searchValue) {
+            query = query + 'where consent_group_key LIKE :term OR vault_export_date LIKE :term '
+        }
+        SQLQuery sqlQuery = getSessionFactory().currentSession.createSQLQuery(query)
+        if (pagination.searchValue) sqlQuery.setString('term', pagination.getLikeTerm())
+        Long totalRows = sqlQuery.uniqueResult()
```
rushtong, line 1493: For code organization, move this to just above the `List<DataUseRestriction> results = sqlQuery` so it's clear where query construction ends and querying begins.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1486,4 +1486,87 @@ class QueryService implements Status {
...
+        new PaginatedResponse(
+                draw: pagination.draw,
+                recordsTotal: totalRows,
+                recordsFiltered: totalRows,
+                data: results,
+                error: ""
+        )
+    }
+
+    Collection findCollectionLinks() {
```
rushtong: Should be a typed collection

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1486,4 +1486,87 @@ class QueryService implements Status {
...
+        SQLQuery sqlQuery = getSessionFactory().currentSession.createSQLQuery(query)
+        List<ConsentCollectionLink> links = sqlQuery.with {
+            addEntity(ConsentCollectionLink)
+            list()
+        }
+        Map<String, SampleCollection> collectionMap = getCollectionIdMap(links)
+        Collection<DataUseRestriction> durs = findDataUseRestrictionByConsentGroupKeyInList(links.collect {
+            it.consentKey
+        })
+        links.each { link ->
```
rushtong: There is some kind of weirdness with respect to how SQLQuery.with types the results it returns. It comes back as an untyped collection, even if the left hand side of the assignment declares the type. If you look around in this class, you'll see that we have to cast the results. I wish there were a better way to do this, but I don't know what that is. In the meantime, this will ensure that links is what you think it is here:
```suggestion
        (links as Collection<ConsentCollectionLink>).each { link ->
```

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1486,4 +1486,87 @@ class QueryService implements Status {
...
+            if (link) {
+                if (link.sampleCollectionId && collectionMap.containsKey(link.sampleCollectionId)) {
```
rushtong: You can probably ignore the nested if condition if you use the null safe operator. 
```suggestion
                if (link?.sampleCollectionId && collectionMap.containsKey(link?.sampleCollectionId)) {
```

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -113,6 +115,20 @@ class UtilityClass {
...
     }
 
+    static void registerConsentCollectionReportMarshaller() {
+        JSON.createNamedConfig(CONSENT_COLLECTION) {
+            it.registerObjectMarshaller( ConsentCollectionLink ) { ConsentCollectionLink link ->
+                return [
+                        id: link.id,
+                        projectKey: link.projectKey,
+                        consentKey: link.consentKey,
+                        sampleCollectionId: link.sampleCollectionId ? link.sampleCollectionId : '',
```
rushtong: Here and below, this can be simplified with Elvis:
```suggestion
                        sampleCollectionId: link.sampleCollectionId ?: '',
```

grails-app/views/base/_topNav.gsp
```diff
@@ -35,7 +35,7 @@
                             <a href="#" class="dropdown-toggle" data-toggle="dropdown">Admin <b class="caret"></b></a>
                             <ul class="dropdown-menu">
                                 <li><a href="${createLink(controller: 'admin', action: 'collectionLinks')}">Consent Collection Links</a></li>
-                                <li><a href="${createLink(controller: 'dataUse', action: 'list')}">Data Use Restrictions</a></li>
+                                <li><a href="${createLink(controller: 'dataUseRestriction', action: 'list')}">Data Use Restrictions</a></li>
```
rushtong: Note that this change means `DataUseController.list` is now unused. We should clean up the dead code in this case.

src/main/webapp/dataUse/DataUseRestriction.js
```diff
@@ -0,0 +1,179 @@
...
+  { text: '500', value: 500 }];
+
+const columns = [
+  {
+    dataField: 'consentGroupKey',
+    text: 'Consent Group',
+    sort: true,
+    formatter: (cell, row, rowIndex, colIndex) =>
+    div({},[
+      a({ href: handleRedirectToConsentGroup(component.serverURL, row.consentGroupKey) },[row.consentGroupKey])
```
rushtong: I'm a little confused about the direct linking here. Does this reload the page at the destination link or is it intercepted by the router? If I understand our use of the router correctly, these should instead be `Link` objects that point to the correct destination.

src/main/webapp/dataUse/SampleCollectionLinks.js
```diff
@@ -0,0 +1,129 @@
...
+  {
+    dataField: 'consentGroupKey',
+    text: 'Consent Group',
+    sort: true,
+    headerStyle: (column, colIndex) => {
+      return { width:  styles.consentCollection.consentKeyWidth};
+    },
+    formatter: (cell, row, rowIndex, colIndex) =>
+      div({}, [
+        a({ href: handleRedirectToConsentGroup(component.serverURL, row.consentGroupKey) }, [row.consentGroupKey])
```
rushtong: Same concern here as in `DataUseRestriction.js`

rushtong (changes requested):
See comments inline.

vvicario:
@rushtong I worked on your feedback, can you please check it? Thanks!

src/main/webapp/dataUse/DataUseRestriction.js
```diff
@@ -0,0 +1,181 @@
...
+      h(Link, {to: {pathname:'/newConsentGroup/main', search: '?consentKey=' + row.consentGroupKey, state: {issueType: 'consent-group', tab: 'documents', consentKey: row.consentGroupKey}}}, [row.consentGroupKey])
+    ])
+  },
+  {
+    dataField: 'id',
+    text: 'View Restrictions',
+    sort: false,
+    formatter: (cell, row, rowIndex, colIndex) =>
+    div({},[
+      a({href: component.serverURL + '/dataUse/show/' + row.id}, ['View Restriction'])
```
rushtong, line 34: Shouldn't this also be a `Link`?

rushtong (approved):
👍 One comment inline

src/main/webapp/dataUse/DataUseRestriction.js
vvicario: I can't use Link  because Data Use Restriction view page has not been migrated yet. I'll do that on US 748

rushtong: 👍

tadeork (approved):
:+1:

PR #232, opened 2019-08-08
Btrx 599 dgil Reimplement BSP Sync

dags:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-599

## Changes
Moved from a "pull" approach to a "push" one. A Broad's internal cron job gets data from BSP service and post it to ORSP in Cloud. 

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/webservice/SampleCollectionController.groovy
```diff
@@ -0,0 +1,87 @@
...
+import groovy.util.logging.Slf4j
+import org.broadinstitute.orsp.SampleCollection
+import org.springframework.web.multipart.MultipartFile
+
+import groovyx.net.http.HttpBuilder
+import org.broadinstitute.orsp.webservice.BspCollection
+
+@Slf4j
+@Resource(readOnly = false, formats = ['JSON', 'CSV'])
+class SampleCollectionController {
```
rushtong: My concern about this is that it is unauthenticated and therefore open to hacking. Since this is open source, it will be easy for anyone to POST to this API any kind of random data.

rushtong (changes requested):
Code looks good. We need some kind of auth scheme for this approach, however.

grails-app/controllers/org/broadinstitute/orsp/webservice/SampleCollectionController.groovy
dags: Yes, I agree. I am looking to decide which approach to use.

grails-app/controllers/org/broadinstitute/orsp/webservice/SampleCollectionController.groovy
```diff
@@ -0,0 +1,92 @@
...
+     * from a cron job running in Broad's internal network, 
+     * finds out new sample collection (not in ORSP db)
+     * and save it on ORSP db.
+     *
+     * @return List of added sample collections
+     */
+    def save() {
+
+        if (sampleCollectionConfiguration.token != request.getHeader('TOKEN')) {
+            log.info("Unauthorized BSP Sync request at: ${new Date().getDateTimeString()}")
```
rushtong, line 42: I think we should be logging all requests for safety, and log as much of the request as possible in the failure case. We will need to follow up on those and will need as much info as possible to diagnose the offending requestor.

rushtong (approved):
One request for additional logging, then 👍

vvicario (approved):

PR #236, opened 2019-08-13
BTRX - 756 - Bug Fixing

LeoForconesi:
## Addresses
[BTRX-756](https://broadinstitute.atlassian.net/browse/BTRX-756): ORSP: Bug Fixing for BTRX-733

## Changes
- Spinner in new sample data cohort
- Spinner in add existing sample data cohort
- UI and styles fixing
- Aprove / Reject buttons in Pending status SDC. For some old projects their status might be null, 

## Testing
-  Spinners for Add New Sample/Data Cohort, and Add Existing Sample/Data Cohort, shoul be visible when they are submitted
- Enter in a Project, resize your browser's window to a smaller size. Project's  (and Sample Data Cohort's) tabs will have now a scroll to move through them, they should have an aligned row.
- There should never be any Sample Data cohort without any action button. 
- Align "On going process" checkbox for date pickers, you can check this when creating a new Sample Data Cohort.

---


src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -86,6 +88,9 @@ class NewConsentGroup extends Component {
       }
     );
   }
+  componentWillUnmount() {
+    spinnerService._unregisterAll();
+  }
```
rushtong, line 94: This is a good example of why the global state approach is so problematic. Imagine that there is a parent component surrounding `NewConsentGroup` that may have other components in it. One of those other components has an active spinner going for some other, unrelated reason. This code will effectively preempt the other component's use of the spinner and now you have introduced a bug.

rushtong (approved):
👍 Looks good, and I'm really hoping the spinner code gets a serious refactor soon.

src/main/webapp/consentGroup/NewConsentGroup.js
LeoForconesi: Thanks for pointing out this, I've used named spinners for `NewConsentGroup` and `LinkWizard` components. Always knowing this will be removed in [BTRX-754](https://broadinstitute.atlassian.net/browse/BTRX-754).

tadeork (commented):
:+1:

PR #237, opened 2019-08-15
BTRX-748 Convert Data Use Restriction View page

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-748

## Changes
Remove dataUse/show.gsp and create a new react component 
## Testing
Go to admin -> Data Use Restrictions -> View Restriction.
View restriction page should be displayed properly

---


grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
```diff
@@ -101,6 +101,8 @@ class UrlMappings {
         '/api/submissions/add-new'(controller: 'submission', action: 'index')
         '/api/submissions/add-new'(controller: 'submission', action: 'save', method: 'POST')
         '/api/data-use/new-restriction'(controller: 'dataUse', action: 'create')
+        '/api/data-use/restriction'(controller: 'dataUse', action: 'findRestriction')
+        '/api/consent/export'(controller: 'dataUse', action: 'exportConsent', method: 'POST')
```
rushtong, line 105: Didn't catch this the first time around, but we have two `DataUseController` classes. This is confusing for both developers and for Grails as it tries to do its own wiring based on naming patterns. Having an API version of `DataUseController` introduced this technical debt so we need to be able to track removing the non-API class. Please reference that ticket here (and create it if it doesn't exist) so it's tracked in the PR.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -1,17 +1,35 @@
...
+    GrailsApplication grailsApplication
+
     def list() {
         render(view: "/mainContainer/index")
     }
-
+    @Override
+    def show() {
+        render(view: "/mainContainer/index", model:[restrictionId: params.id])
+    }
```
rushtong, line 33: nit: add an empty line between methods.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -1,17 +1,35 @@
 import grails.converters.JSON
+import grails.core.GrailsApplication
 import grails.rest.Resource;
 import groovy.util.logging.Slf4j;
```
rushtong: No need for semi-colons, here and at the end of the imports.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -1,17 +1,35 @@
 import grails.converters.JSON
```
rushtong, line 3: This class is missing a package name. Add this at the top of the file:
```
package org.broadinstitute.orsp.api

```

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -1,17 +1,35 @@
...
+import org.broadinstitute.orsp.utils.UtilityClass
 import org.broadinstitute.orsp.webservice.PaginationParams;
 
 @Slf4j
 @Resource
 class DataUseController extends AuthenticatedController {
 
+    ConsentService consentService
+    ConsentExportService consentExportService
+    GrailsApplication grailsApplication
```
rushtong: This is already in `AuthenticatedController` - there's no need to instantiate another one here that overrides the one in the superclass.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -23,4 +41,56 @@ class DataUseController extends AuthenticatedController {
...
+        UtilityClass.registerSampleCollectionMarshaller()
+        DataUseRestriction restriction = DataUseRestriction.findById(params.id)
+        List<String> summary = consentService.getSummary(restriction)
+        Issue consent = queryService.findByKey(restriction.consentGroupKey)
+        String consentGroupName = consent.summary
+        ConsentResource consentResource = consentService.buildConsentResource(restriction, consentGroupName)
+        Collection<ConsentCollectionLink> collectionLinks = queryService.findCollectionLinksByConsentKey(restriction.consentGroupKey)
+        collectionLinks.retainAll { it.sampleCollectionId != null }
+        Set<SampleCollection> samples = collectionLinks.collect{ it.sampleCollection }
+        JSON.use(UtilityClass.SAMPLES) {
```
rushtong, line 56: This is a very confusing closure. As a developer reading this, why would one use `SAMPLES` as a model for rendering a `DataUseRestriction`? Following the code, this makes a little more sense in that it seems to apply only to the `collectionLinks` and not to everything else being rendered. 

Even then, why does `SAMPLES` apply to `collectionLinks`? 

I think the naming here needs to be cleaned up to be more legible.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -23,4 +41,56 @@ class DataUseController extends AuthenticatedController {
...
+            render([restriction    : restriction,
+                    restrictionUrl : restriction.getConsentUrl(getConsentServiceUrl()),
+                    summary        : summary,
+                    consent        : consent,
+                    collectionLinks: samples,
```
rushtong: Huh? `collectionLinks` are `samples`?

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -122,6 +123,17 @@ class UtilityClass {
         }
     }
 
+    static void registerSampleCollectionMarshaller() {
+        JSON.createNamedConfig(SAMPLES) {
+            it.registerObjectMarshaller( ConsentCollectionLink ) { ConsentCollectionLink sc ->
```
rushtong: This should be looking at a `SampleCollection`, not a `ConsentCollectionLink`. If the usages of this are passing in a CCL, then that's a category mistake - usages should be passing in a `SampleCollection`.

package-lock.json
```diff
@@ -0,0 +1,14173 @@
+{
```
rushtong: What happened to this file in feature? Why is it being re-added here? This should be in a separate PR as it is not related to the ticket.

src/main/webapp/dataUse/DataUseRestrictionDetails.js
```diff
@@ -0,0 +1,263 @@
...
+      },
+      samples: []
+    };
+    this.exportConsent = this.exportConsent.bind(this);
+  }
+
+  componentDidMount() {
+    spinnerService.show(DUR_SPINNER);
+    this.init();
+  }
```
rushtong, line 87: nit: line break between methods

src/main/webapp/main/Routes.js
```diff
@@ -2,9 +2,10 @@ import React from "react";
 import { Switch, Route } from "react-router-dom";
 import RolesManagement from "../rolesManagement/RolesManagement";
 import ReviewCategories from "../reviewCategories/ReviewCategories";
-import Profile from "../profile/Profile";
 import Main from "./Main";
 import FundingsSourceReport from "../fundingSourceReport/FundingsSourceReport";
+import DataUseRestrictionDetails from "../dataUse/DataUseRestrictionDetails";
+import Profile from "../profile/Profile";
```
rushtong, line 8: nit: We should get in the habit of optimizing imports in JS files which will follow the eslint rules. When you do that, it sorts them appropriately (which actually uses paths, not class names) so we aren't moving lines around for no reason.
```
import React from 'react'
import { Route, Switch } from 'react-router-dom'
import AboutPage from '../aboutPage/AboutPage'
import NewConsentGroup from '../consentGroup/NewConsentGroup'
import DataUseRestrictionDetails from '../dataUse/DataUseRestrictionDetails'
import DataUseRestrictionIndex from '../dataUse/DataUseRestrictionIndex'
import DataUseLetterIndex from '../dataUseLetter/DataUseLetterIndex'
import FundingsSourceReport from '../fundingSourceReport/FundingsSourceReport'
import { LinkWizard } from '../linkWizard/LinkWizard'
import Profile from '../profile/Profile'
import ReviewCategories from '../reviewCategories/ReviewCategories'
import RolesManagement from '../rolesManagement/RolesManagement'
import Main from './Main'

```

rushtong (changes requested):
See comments inline

grails-app/controllers/org/broadinstitute/orsp/UrlMappings.groovy
vvicario: I agree, I created story:  https://broadinstitute.atlassian.net/browse/BTRX-765

vvicario:
@rushtong Please when you have time can you check it? Thanks

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -23,4 +43,56 @@ class DataUseController extends AuthenticatedController {
...
+        ConsentResource consentResource = consentService.buildConsentResource(restriction, consentGroupName)
+        Collection<ConsentCollectionLink> collectionLinks = queryService.findCollectionLinksByConsentKey(restriction.consentGroupKey)
+        collectionLinks.retainAll { it.sampleCollectionId != null }
+        Set<SampleCollection> samples = collectionLinks.collect{ it.sampleCollection }
+        JSON.use(UtilityClass.SAMPLES) {
+            render([restriction    : restriction,
+                    restrictionUrl : restriction.getConsentUrl(getConsentServiceUrl()),
+                    summary        : summary,
+                    consent        : consent,
+                    samples: samples,
```
rushtong: nit: alignment - if you want to align the other arguments, this one should be too.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -23,4 +43,56 @@ class DataUseController extends AuthenticatedController {
...
+                    consent.summary)
+            response.status = 200
+            render([message: "Consent successfully exported to DUOS: $resource.name"] as JSON)
+        } catch (ConsentException e) {
+            response.status = 500
+            render(status: 500, text: e as JSON, contentType: 'application/json')
+        } catch (Exception e) {
+            response.status = 500
+            render(status: 500, text: e as JSON, contentType: 'application/json')
+        }
```
rushtong, line 89: These two exception cases do the exact same thing. Should be collapsed into a single catch.

rushtong (approved):
Minor comments inline, but 👍

tadeork (commented):
:+1:

tadeork (approved):

PR #238, opened 2019-08-15
[BTRX-744] Session timeout 

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-744

## Changes
Redirects to index when user lost session.

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/ApiInterceptor.groovy
```diff
@@ -42,7 +42,7 @@ class ApiInterceptor implements UserInfo  {
 
     def forbidden() {
         response.status = 403
-        render([error: 'access denied'] as JSON)
+        render(view: "/index")
```
rushtong: This doesn't seem like an appropriate action for `forbidden`. There are a host of urls that we direct here which are phishing attempts. I think the solution to this ticket will be how it is that this endpoint is being called instead of changing what the response is. Better to handle the case where auth session is checked and if no session is available, then redirect to the index.

rushtong (changes requested):
See inline comment.

grails-app/controllers/org/broadinstitute/orsp/ApiInterceptor.groovy
tadeork: I'm going to move this to the ``before`` method and will remove the ``forbidden`` method because the forbidden method is already implemented in ``ReadOnlyInterceptor``. In the other hand the phishing attempts are handled in the ``ErrorController``. Until we separate the front from the back I think we need to redirect to index but not inside the forbidden method, like you stated in your comment.

rushtong (approved):
👍 Much cleaner, thank you for the changes!

vvicario (approved):

PR #241, opened 2019-08-16
[BTRX-742] Convert submission form

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-742

## Changes
Submission form migrated to reactjs, removed old views.

## Testing

---


grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -55,48 +48,61 @@ class SubmissionController extends AuthenticatedController {
...
-        if (params?.submissionId) {
-            submission = Submission.findById(params?.submissionId)
-            submission.number = Integer.parseInt(params?.submissionNumber)
-            submission.type = params?.submissionType
-            submission.comments = params?.comments
-            if (!submission.save(flush: true)) {
-                flash.message = submission.errors.allErrors.collect { it }.join("<br/>")
+        JsonParser parser = new JsonParser()
+        JsonArray dataSubmission = parser.parse(request.parameterMap["submission"].toString())
+        List<MultipartFile> files = request.multiFileMap.collect { it.value }.flatten()
```
rushtong: Here and above, you might want to use the null-safe operator (`?`) to traverse some of these objects that might not really be part of the request.

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -55,48 +48,61 @@ class SubmissionController extends AuthenticatedController {
...
-        Submission submission
-        if (params?.submissionId) {
-            submission = Submission.findById(params?.submissionId)
-            submission.number = Integer.parseInt(params?.submissionNumber)
-            submission.type = params?.submissionType
-            submission.comments = params?.comments
-            if (!submission.save(flush: true)) {
-                flash.message = submission.errors.allErrors.collect { it }.join("<br/>")
+        JsonParser parser = new JsonParser()
+        JsonArray dataSubmission = parser.parse(request.parameterMap["submission"].toString())
```
rushtong: `parser.parse` returns a `JsonElement`, not a `JsonArray`, so you need to cast it. Something along these lines:
```
        JsonElement el = parser.parse(request.parameterMap["submission"].toString())
        if (el.jsonArray) {
            JsonArray dataSubmission = el.asJsonArray
        }
```

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
rushtong: I'm not sure about how you're using request here. This is an instance of `HttpServletRequest` which doesn't have a `multiFileMap` field/method. What this should be doing under the hood is using a `GrailsWebRequest` which knows if it's a `multipartRequest` request or not. If it is a multipartRequest, then Grails will store all of the elements of `multiFileMap` as elements of the parameterMap. I'm looking into a better way to do this at the moment, but the cast here is concerning. There might be an easy way to implement a trait that exposes this more readily to the controller. I looked at implementing `WebAttributes` but that wasn't very helpful, even though it gives you direct access to the `GrailsWebRequest`

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -55,48 +48,61 @@ class SubmissionController extends AuthenticatedController {
...
-        } else {
-            submission = new Submission(
-                    projectKey: params.projectKey,
-                    number: Integer.parseInt(params?.submissionNumber),
-                    author: getUser()?.displayName,
-                    type: params?.submissionType,
-                    comments: params.comments,
-                    createDate: new Date(),
-                    documents: new ArrayList<StorageDocument>())
+            if(!files?.isEmpty()) {
```
rushtong: Even though you're looking at files here, the thing you're using is the `parts` part of the `request`. You might be able to iterate over `request.parts` instead.

package-lock.json
```diff
@@ -0,0 +1,14207 @@
+{
```
rushtong: I mentioned this in Veronica's PR too - this should be in a separate PR, not this one - it's unrelated.

src/main/webapp/main/Routes.js
```diff
@@ -17,13 +18,16 @@ const Routes = ( props ) => (
...
+    <Route path= {"/report/reviewCategories"} render = {(routeProps) =>  <ReviewCategories {...routeProps} {...props}/> }/>
     <Route path= {"/index/profile"} render = {(routeProps) =>  <Profile {...routeProps} {...props}/> }/>
     <Route path= {"/index/about"} render = {(routeProps) =>  <AboutPage {...routeProps} {...props}/> }/>
     <Route path= {"/dataUse/list"} render = {(routeProps) =>  <DataUseRestrictionIndex {...routeProps} {...props}/> }/>
     <Route path= {"/consent-group/use-existing"} render = {(routeProps) => <LinkWizard {...routeProps} {...props}/> }/>
     <Route path= {"/consent-group/new"} render = {(routeProps) =>  <NewConsentGroup {...routeProps} {...props}/> }/>
     <Route path= {"/dataUseLetter/show"} render = {(routeProps) =>  <DataUseLetterIndex {...routeProps} {...props}/> }/>
+    <Route path= {"/submissions/add-new"} render={(routeProps) => <SubmissionForm {...routeProps} {...props} />} />
+
+
```
rushtong: nit: rm extra lines here.

src/main/webapp/util/UrlConstants.js
```diff
@@ -66,12 +66,17 @@ export const UrlConstants = {
   infoLinkUrl: context + '/api/info-link',
   restrictionUrl: context + '/api/dur',
 
+  submissionsUrl: context + '/api/submissions',
+  submissionsAddNewUrl: '/submissions/add-new',
```
rushtong, line 71: Does this not need a context too?

rushtong (changes requested):
See comments inline. I did not test the UI functionality, just looking at code here.

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
tadeork: This seems not be working for iteration.

rushtong: I don't know what the exact right answer is here, only that accessing `request.multiFileMap` directly is brittle.

src/main/webapp/util/UrlConstants.js
tadeork: No because this url is used to go through the router component.

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
tadeork: OK... I'll try some ideas.

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
tadeork: Should I try to implement a trait with something like this inside?
```MultiValueMap<String, Object> map = new LinkedMultiValueMap<>();
List<ByteArrayResource> files = new ArrayList<>();

for(MultipartFile file : multiPartFileList) {
files.add(new ByteArrayResource(file.getBytes()));
}
map.put("files", files);

Comment (rushtong) [reply grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy]: OK, there are two paths we can go down here. 

### MultiFileMap
In the master branch, we have a method to add a document to an issue that looks a lot like what you have in this PR. It's older code, but functional, so here is that version:
```
        try {
            if (request.multiFileMap?.files) {
                storageProviderService.saveMultipartFiles(
                        (String) getUser()?.displayName,
                        (String) getUser()?.userName,
                        (String) issue.projectKey,
                        (String) params.type,
                        (Collection<MultipartFile>) request.multiFileMap.files)
            }
            redirect(controller: 'irb', action: "show", params: [id: issue.projectKey, tab: 'documents'])
        } catch (Exception e) {
            render([status: 500, text: [error: e.message] as JSON])
        }
```

### DiskFileItem
In the original submission controller, we had a separate method for adding a file. This is newer code, and a slightly more preferable approach. What Grails does behind the scenes is taking a multipart request and mapping the file elements to the parameter map. That gives us `params.files` which may or may not exist:
```
            DiskFileItem file = params?.files?.part?.fileItem
            if (file) {
                StorageDocument document = storageProviderService.saveFileItem(
                        (String) getUser()?.displayName,
                        (String) getUser()?.userName,
                        (String) params?.projectKey,
                        (String) params?.type,
                        file
                )
                ....
             } else {
                flash.error = "Please select a file to upload"
            }
```

rushtong (approved):
Thank you for the changes!

vvicario (approved):

PR #245, opened 2019-08-28
BTRX - 615 - Convert QA Event Report

LeoForconesi:
## Addresses
[BTRX-615](https://broadinstitute.atlassian.net/browse/BTRX-615): ORSP: Convert QA Event Report

## Changes
- Relocate TextEditor component to Components folder
- Update statusEvent Service and controller to return neccessary data to the Ui. 
- Add to Table component a new button to export Pdf format
- Add the ability to hide columns from Excel export through an array of strings
- Modify MultiTab component so it can detect if a default tab is required without having to re-render the component containing MultiTab.

-**Note:** The functionality to export the entire data from tables (pdf, excel, csv, print) will be handled all thogheter in [BTRX-746](https://broadinstitute.atlassian.net/browse/BTRX-746).
- Project Report page will be converted as part of [BTRX-773](https://broadinstitute.atlassian.net/browse/BTRX-773)

## Testing
- Access to 'QA Event Report' option from the top navigation bar
- Test Filters. Make combinations with this filter, adding only after date, before date, project type, then their possible combinations. After a search is made, if only IRB or Non-IRB type projects are throwed as a filter result, only its tab should be visible.
- Clear button should clean filter box.
- Each Tab has their own pagination, and should mantain its possition if you change between tabs.
- Check this logic for Age column :
> The first event to calculate time from is either "Reviewing Form" if it exists,
> or the very first event if it does not.
> For IRBs, we don't need to be picky about the initial status because they are not
> specifically required for QA reporting, only NE/NHSR projects.
> Duration is calculated as this event's creation date compared to the next event's creation date.
> If there isn't a next event, then we use today's date to calculate the duration which indicates
> how long the project has been left in that final state.
> Terminal Event Case:
> If the event is a terminal event (i.e. completed, abandoned, etc.) then the duration is
> calculated from the first event to the terminal state event.

---


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -650,23 +649,75 @@ class QueryService implements Status {
...
-        if (options.issueTypeNames) {
-            typeList = options.issueTypeNames
-        } else {
-            typeList = [IssueType.NE.name, IssueType.NHSR.name, IssueType.IRB.name]
+    @SuppressWarnings(["GroovyAssignabilityCheck"])
+    PaginatedResponse findIssuesForStatusReport(PaginationParams paginationOptions, QueryOptions queryOptions) {
+        String orderColumn = statusReportPaginationOrder(paginationOptions)
+        SessionFactory sessionFactory = grailsApplication.getMainContext().getBean('sessionFactory')
+        final session = sessionFactory.currentSession
+        final StringBuffer query = new StringBuffer(' select distinct i.project_key from issue i left outer join issue_extra_property ie on ie.issue_id = i.id where i.deleted = 0 and i.type != "Consent Group"')
```
vvicario: please use constant instead of  "Consent Group"

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -68,20 +65,50 @@ class StatusEventService {
                     } else {
                         duration = new Period(event.created.time, new Date().time, PeriodType.yearMonthDay())
                     }
-                    new StatusEventDTO(event, duration)
+                    new StatusEventDTO(new Issue(), event, duration)
```
vvicario: It's not necessary to create a new issue, send only event and duration

src/main/webapp/components/MultiTab.js
```diff
@@ -1,26 +1,34 @@
...
-      key: ""
+      key: "",
+      defaultActive: ""
     };
 
     this.handleSelect = this.handleSelect.bind(this);
   }
 
- handleSelect(key) {
+  handleSelect(key) {
```
tadeork: nit: spacing

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
vvicario: It's not necessary to create a new issue, send only event and duration

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -36,18 +36,15 @@ class StatusEventService {
         results as Collection<Event>
     }
 
-    List<StatusEventDTO> getStatusEventDTOs(String projectId) {
-        List<Event> statusEvents = getStatusEventsForProject(projectId).sort {
-            a, b -> a.created <=> b.created
-        }
-        if (statusEvents.isEmpty()) return Collections.emptyList()
+    List<StatusEventDTO> getStatusEventDTOs(List<Event> statusEvents) {
+        if (statusEvents && statusEvents.isEmpty()) return Collections.emptyList()
```
vvicario: CollectionUtils.isEmpty could be used to validate statusEvents

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -68,20 +65,50 @@ class StatusEventService {
...
+                statusEventDTO?.setIssue(issue)
+                result.add(statusEventDTO)
+            } else {
+                result.add(new StatusEventDTO(issue, null, null))
+            }
+        }
+        result
+    }
+
     class StatusEventDTO {
```
vvicario: I think this could be in a separate class

src/main/webapp/components/MultiTab.js
```diff
@@ -1,26 +1,34 @@
...
   componentDidMount() {
-    this.setState({key: this.props.defaultActive});
+    this.setState({key: this.props.defaultActive, defaultActive: this.props.defaultActive});
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps.defaultActive && nextProps.defaultActive !== prevState.defaultActive){
+      return { key: nextProps.defaultActive, defaultActive: nextProps.defaultActive};
+    }
+    else return null;
```
tadeork: I think it's better with brackets
```suggestion
    else {
       return null;
     }
```

src/main/webapp/components/TableComponent.js
```diff
@@ -39,15 +39,21 @@ export const TableComponent = hh(class TableComponent extends Component {
                     fileName={this.props.fileName}
                     fileType={EXPORT_FILE.XLSX.mimeType}
                     fileExtension={EXPORT_FILE.XLSX.extension}
+                    hide={this.props.hideColumns !== undefined && this.props.hideColumns.length > 0 ? this.props.hideColumns : []}
```
tadeork: Why this validation ends with an empty array?

src/main/webapp/qaReport/FilterPanel.js
```diff
@@ -0,0 +1,89 @@
...
+    super(props);
+    this.state = {};
+  }
+
+  render() {
+    return(
+      div({
+      },[
+        Panel({ title: "Filter Projects" }, [
+          div({className: "row"}, [
```
tadeork, line 39: Probably we should create a functional component for rows and use it with `.map()`

src/main/webapp/qaReport/QaReport.js
```diff
@@ -0,0 +1,269 @@
...
+    }) ;
+  };
+
+  applyFilterPanel = async () => {
+    if (this.state.projectType.value === 'irb') {
+      this.setState(prev => {
+        prev.activeTab = IRB;
+        prev[NO_IRB].hide = true;
+        }, async () =>
+        await this.tableHandler(0, this.state[IRB].sizePerPage, this.state[IRB].search, this.state[IRB].sort, this.state[IRB].currentPage, IRB)
```
tadeork: I think it would be more readable if this can be split in multiple lines.
```suggestion
        await this.tableHandler(
          0, 
          this.state[IRB].sizePerPage, 
          this.state[IRB].search, 
          this.state[IRB].sort, 
          this.state[IRB].currentPage, 
          IRB
        )
      );
```

src/main/webapp/util/QaReportConstants.js
```diff
@@ -0,0 +1,55 @@
...
+
+export const QA_EVENT_SORT_NAME_INDEX = {
+  'projectKey': 0,
+  'approvalStatus': 1,
+  'age': 2,
+  'assignee': 3,
+  'type' : 4
+};
+
+export const COLUMNS_TO_HIDE_FROM_EXCEL = ['attachments', 'reporter', 'requestDate', 'reviewCategory', 'summary', 'issueStatus'];
```
tadeork: nit: EOF.

tadeork (commented):
Just some minor comments. It's OK in general.

tadeork (approved):
Front-end code it's OK to me.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -650,23 +649,29 @@ class QueryService implements Status {
...
+    @SuppressWarnings(["GroovyAssignabilityCheck"])
+    List<StatusEventDTO> findIssuesForStatusReport(Collection<String> issueTypeNames) {
+        if (issueTypeNames.isEmpty()) {
+            log.error("An error has occurred trying to find Issue Types from empty list for Quality Assurance Report.")
+            throw new IllegalArgumentException("Error trying to fetch Projects by unspecified type.")
         }
+        SessionFactory sessionFactory = grailsApplication.getMainContext().getBean('sessionFactory')
+        final session = sessionFactory.currentSession
+        final StringBuffer query = new StringBuffer(' select distinct * from issue i where i.deleted = 0 and i.type != :consentGroup')
+        query.append(' and i.type IN :filterType ')
```
rushtong: These two lines should just be a `String`

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -56,69 +61,50 @@ class StatusEventService {
         // Terminal Event Case:
         // If the event is a terminal event (i.e. completed, abandoned, etc.) then the duration is
         // calculated from the first event to the terminal state event.
-        List<StatusEventDTO> eventDTOs = statusEvents.
-                withIndex().
+        List<StatusEventDTO> eventDTOs = statusEvents?.
+                withIndex()?.
```
rushtong: Not sure you need the `?`  in `withIndex` ... that's a built-in method on the groovy `Collection` class. Using elvis on `statusEvents` makes sense. But if it exists, then it has `withIndex` innately.

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -56,69 +61,50 @@ class StatusEventService {
...
                 collect { Event event, int i ->
-                    Period duration
+                    Long durationTime
```
rushtong, line 67: I'm curious about why this is better than `Period`. It could be more efficient, I just don't know. I do remember that `Period` handled some date conversions for me that were problematic at the time, but I don't remember what those were.

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
```diff
@@ -56,69 +61,50 @@ class StatusEventService {
...
-                    break
-                case EventType.ABANDON:
-                    6
-                    break
-                default:
-                    0
-                    break
+        Map<String, List<Event>> eventMap = eventList.groupBy {it.projectKey}
+        List<StatusEventDTO> result = new ArrayList<>()
+        issues.collect { issue ->
```
rushtong: This returns a collection that you're not using. Probably better to just use `each` in this case.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -83,7 +85,27 @@ class UtilityClass {
...
+
+    static void registerQaReportIssueMarshaller() {
+        JSON.createNamedConfig(ISSUE_FOR_QA) {
+            it.registerObjectMarshaller( StatusEventDTO ) { StatusEventDTO statusEvent ->
+                return [
+                    id             : statusEvent.issue.id,
+                    type           : statusEvent.issue.type,
+                    projectKey     : statusEvent.issue.projectKey,
+                    summary        : statusEvent.issue.summary,
+                    status         : statusEvent.issue.approvalStatus == IssueStatus.Legacy.name ? statusEvent.issue.status : statusEvent.issue.approvalStatus,
```
rushtong, line 103: The fact that this ternary needs to exist is concerning. We have two different statuses that are sort of the same thing ... sometimes. This is super confusing for any dev to have to reason about and will cause some bugs as folks try to remember which status to show in which case. Instead, we should have a single place where we get the status that does this kind of check for us. You can override this in `Issue` with this logic with a nice comment explaining why this problem exists ... and hopefully a link to the ticket that fixes this across the board.

src/main/webapp/components/MultiTab.js
```diff
@@ -1,26 +1,39 @@
...
+      this.props.handleSelect(key);
+    }
   }
 
   componentDidMount() {
-    this.setState({key: this.props.defaultActive});
+    this.setState({key: this.props.defaultActive, defaultActive: this.props.defaultActive});
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
dags: This class seems to be a good candidate to be refactored as a functional component, specially to get rid of getDerivedStateFromProps. Not required now but may be a good exercise.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
wloforteBroad: @rushtong  I totally agree with you and we just discuss this with Leo. We will create a ticket to add this logic in `Issue` (and also add a comment with the explanation of why we need this) and delete this ternary from all the places we are using it. Sounds good?

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
LeoForconesi: As it was before, it used Period as way to build and return the string "xx Years and nn months and yy days" directly from groovy method.
Since we need to sort this in bootstrap table, it was easier to return milliseconds and parse each value as string using javascript Date.
If you consider it would be better to use Java 8 object i could use `ChronoUnit.between` method. [Doc](https://docs.oracle.com/javase/tutorial/datetime/iso/period.html).

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
LeoForconesi: Changed to `each`.

src/main/webapp/components/MultiTab.js
LeoForconesi: This change might modify how it is used in other components, and would require re-test the branch. I can implement this in the branch i'm currently working on. Let me know if this sounds good for you.

grails-app/services/org/broadinstitute/orsp/StatusEventService.groovy
rushtong: Thanks, makes sense. I'm fine with the current approach.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
rushtong: Sounds good, thank you!

wloforteBroad: @rushtong  https://broadinstitute.atlassian.net/browse/BTRX-776

src/main/groovy/org/broadinstitute/orsp/StatusEventDTO.groovy
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.orsp
+
+import org.joda.time.Period
```
dags: not used

src/main/webapp/components/MultiTab.js
```diff
@@ -1,26 +1,39 @@
...
+      defaultActive: ""
     };
 
     this.handleSelect = this.handleSelect.bind(this);
   }
 
- handleSelect(key) {
+  handleSelect(key) {
     this.setState({ key: key });
+    if (this.props.handleSelect !== undefined) {
```
dags: != null  verifies both null and undefined. Check if applies here ..

grails-app/controllers/org/broadinstitute/orsp/StatusEventController.groovy
```diff
@@ -1,23 +1,26 @@
...
 import org.joda.time.Period
 
-import java.text.DateFormat
-import java.text.ParseException
 
 @Slf4j
+@Resource(readOnly = false, formats = ['JSON'])
 class StatusEventController extends AuthenticatedController {
+    final static String NO_IRB = "noIrb"
+    final static String ALL_PROJECTS = 'all'
```
rushtong: Unused?

grails-app/controllers/org/broadinstitute/orsp/StatusEventController.groovy
```diff
@@ -1,23 +1,26 @@
...
 
-    private Map<IssueType, Collection<Issue>> getGroupedIssues(QueryOptions options) {
-        queryService.
-                findIssuesForStatusReport(options).
-                groupBy { IssueType.valueOfName(it.type) }
+    def qaEventReport() {
+        render(view: "/mainContainer/index")
     }
 
+    // TODO: unused, this must be re-implemented
```
rushtong: If this really is unused, remove.

grails-app/controllers/org/broadinstitute/orsp/StatusEventController.groovy
```diff
@@ -27,44 +30,28 @@ class StatusEventController extends AuthenticatedController {
...
-        }
-        Map<IssueType, Collection<Issue>> groupedIssues = getGroupedIssues(options)
-        Collection<Issue> irbs = groupedIssues.getOrDefault(IssueType.IRB, Collections.emptyList())
-        Collection<Issue> nonIrbs = groupedIssues.getOrDefault(IssueType.NE, Collections.emptyList()) +
-                groupedIssues.getOrDefault(IssueType.NHSR, Collections.emptyList())
-        Map<String, Period> periodMap = calculateIssuePeriods(irbs + nonIrbs)
-        [irbs: irbs, nonIrbs: nonIrbs, periodMap: periodMap]
-    }
-
+    // TODO: unused, this must be re-implemented
```
rushtong: Ditto

grails-app/controllers/org/broadinstitute/orsp/StatusEventController.groovy
```diff
@@ -27,44 +30,28 @@ class StatusEventController extends AuthenticatedController {
...
         [statusEvents: eventDTOs, issue: issue]
     }
 
+    def findQaEventReport() {
+        UtilityClass.registerQaReportIssueMarshaller()
+        Collection<String> issueTypeNames = new ArrayList<String>()
+
+        if (params.tab) {
+            issueTypeNames = params.tab == NO_IRB ? [IssueType.NE.name, IssueType.NHSR.name] : [IssueType.IRB.name]
+        }
```
rushtong: Another way to avoid mutation is to instantiate the collection in the conditional. Something like (I used EnumSet just for flavor):
```
        Collection<String> issueTypeNames = (params?.tab == NO_IRB) ? 
                EnumSet.of(IssueType.NE, IssueType.NHSR)*.getName() : 
                EnumSet.of(IssueType.IRB)*.getName() 
```

src/main/webapp/components/TableComponent.js
rushtong: Same question, this looks confusing.

src/main/webapp/qaReport/NoIrbTable.js
```diff
@@ -0,0 +1,47 @@
+import { Component } from 'react';
+import { TableComponent } from "../components/TableComponent";
+import {
+  columns,
+  COLUMNS_TO_HIDE_FROM_EXCEL,
+  defaultSorted,
+  NO_IRB,
+  SIZE_PER_PAGE_LIST
+} from "../util/QaReportConstants";
```
rushtong: nit: indentation is different from `IrbTable.js`. I suggest you pick one and be consistent.

src/main/webapp/qaReport/QaReport.js
```diff
@@ -0,0 +1,237 @@
...
+
+  async componentDidMount() {
+    await this.init();
+  }
+
+  async init() {
+    await this.tableHandler(IRB);
+  }
+
+  tableHandler = async ( tab) => {
```
rushtong: nit spacing:
```suggestion
  tableHandler = async (tab) => {
```

src/main/webapp/util/Utils.js
```diff
@@ -53,11 +53,26 @@ export const buildUrlToConsentGroup = (serverURL, consentKey, projectKey) => {
...
+  let hours = Math.floor(mins/60);
+  let days = Math.floor(hours/24);
+  let months = Math.floor(days/31);
+  let years = Math.floor(months/12);
+  months=Math.floor(months%12);
+  days = Math.floor(days%31);
+  hours = Math.floor(hours%24);
+  mins = Math.floor(mins%60);
+  secs = Math.floor(secs%60);
+  return { days: days, months: months, years: years, hours: hours, mins: mins, secs: secs } ;
```
rushtong: Have you looked at the moment library? I was using it in the old javascript code and it is a lot safer for doing this kind of work.
https://www.npmjs.com/package/react-moment
https://momentjs.com/docs/

rushtong (commented):
Looks good, could use a little cleanup. See comments inline.

grails-app/controllers/org/broadinstitute/orsp/StatusEventController.groovy
LeoForconesi: Cool, taking note. Thanks for the recommendation.

src/main/webapp/components/TableComponent.js
LeoForconesi: I've removed this condition. The idea was that if this prop wasn't included on the component, it wouldn't fail when iterating an undefined array, but at the same time no excel column would be hidden. Also I've changed its variable name from `hideColumns` to `hideXlsxColumns`.

src/main/webapp/util/Utils.js
LeoForconesi: I didn't thought of it, it looks interesting! Is it ok if I implement this on my next PR? I'm working on the continuation of this page on this story -> [BTRX-773](https://broadinstitute.atlassian.net/browse/BTRX-773). I just want to avoid retesting and advance from this huge PR if its ok for you @rushtong.

LeoForconesi: I've implemented https://momentjs.com/docs/ moment.duration() it works! i'll leave the react-moment to the other story.

rushtong (approved):
👍

PR #246, opened 2019-08-28
ORSP: Convert Data Use Restriction Edit

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-749

## Changes
Added new component for data use restriction
## Testing
Go to Consent Group -> Documents -> Upload a document with type Data Use Limitation -> Create Restriction. Restriction should be created.
After creation try to edit.
---


grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -89,6 +94,19 @@ class DataUseController extends AuthenticatedController {
         }
     }
 
+    def createSdul() {
+        try {
+            DataUseRestriction restriction = DataUseRestriction.findByConsentGroupKey(request.JSON.consentGroupKey)
+            restriction = DataUseRestrictionParser.fromParams(restriction, request.JSON)
+            dataUseLetterService.createSdul(restriction, getUser()?.displayName)
+            response.status = 200
```
dags, line 88: Should be 201 ?

src/main/webapp/components/Documents.js
```diff
@@ -11,9 +11,18 @@ import './Documents.css';
 import { UrlConstants } from "../util/UrlConstants";
 import { spinnerService } from "../util/spinner-service";
 import { Spinner } from "./Spinner";
+import { Link } from 'react-router-dom';
+import '../components/Btn.css';
```
dags: Do we still need this css ?

src/main/webapp/components/MultiSelect.js
```diff
@@ -87,6 +87,7 @@ export const MultiSelect = hh(class MultiSelect extends Component {
       }, [
           div({ className: "inputFieldSelectWrapper" }, [
             h(AsyncSelect, {
+              key:this.props.id,
```
dags: I think this is not required.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -89,6 +94,19 @@ class DataUseController extends AuthenticatedController {
         }
     }
 
+    def createSdul() {
+        try {
+            DataUseRestriction restriction = DataUseRestriction.findByConsentGroupKey(request.JSON.consentGroupKey)
+            restriction = DataUseRestrictionParser.fromParams(restriction, request.JSON)
```
rushtong: These two construction lines are a little confusing. `request.JSON.consentGroupKey` can fail if JSON doesn't have the right property. If it isn't found, this should probably return a 404. Secondly, it's strange to see a find immediately followed by a reassignment. I can follow the code to see that this is getting the original restriction, then updating it from the passed in parameters. What this should look like is a dedicated service method that `finds and updates a restriction from parameters` which then returns the mutated/unsaved restriction.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
```diff
@@ -10,9 +10,9 @@ class DataUseRestrictionParser {
         dataUseRestriction = dataUseRestriction == null ? new DataUseRestriction() : dataUseRestriction
         dataUseRestriction.consentGroupKey = params.consentGroupKey
         dataUseRestriction.consentPIName = params.consentPIName
-        dataUseRestriction.generalUse = IssueUtils.getBooleanForParam(params.generalUse)
-        dataUseRestriction.hmbResearch = IssueUtils.getBooleanForParam(params.hmbResearch)
-        dataUseRestriction.manualReview = IssueUtils.getBooleanForParam(params.manualReview)
+        dataUseRestriction.generalUse = params.generalUse instanceof Boolean ? params.generalUse : IssueUtils.getBooleanFromString(params.generalUse)
+        dataUseRestriction.hmbResearch = params.hmbResearch instanceof Boolean ? params.hmbResearch : IssueUtils.getBooleanFromString(params.hmbResearch)
+        dataUseRestriction.manualReview = params.manualReview instanceof Boolean ? params.manualReview : IssueUtils.getBooleanFromString(params.manualReview)
```
rushtong: Pretty sure this looks like an over-complication of something that should be very simple. Instead of making all of these changes here, why not just handle the `instanceof` in `IssueUtils.getBooleanForParam`?

grails-app/utils/org/broadinstitute/orsp/utils/IssueUtils.groovy
```diff
@@ -103,7 +103,7 @@ final class IssueUtils {
 
     }
 
-    static Boolean getBooleanForParam(String param) {
+    static Boolean getBooleanFromString(String param) {
```
rushtong: Suggestion: revert the name change and add the boolean check first, i.e. add a line above the two ifs:
```
if (param instanceof Boolean) return param
```
That one line saves you 15 unnecessary edits.

src/main/webapp/components/Documents.js
rushtong: We should be asking that question for all CSS. If we can inline style something, do it. Shared CSS is a huge headache. We should try to limit any css classes to the standard bootstrap stuff we use.

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+import { h1, h, div, hh, h2, strong, ul, i, p, li, label, abbr, input, span, hr, button } from 'react-hyperscript-helpers';
+import { InputFieldText } from "../components/InputFieldText";
+import { InputYesNo } from "../components/InputYesNo";
+import { Search } from "../util/ajax";
+import { InputFieldTextArea } from "../components/InputFieldTextArea";
+import { MultiSelect } from "../components/MultiSelect";
+import { AlertMessage } from '../components/AlertMessage';
+import { DataUse, ConsentGroup } from "../util/ajax";
+import { isEmpty } from '../util/Utils';
+import { UrlConstants } from '../util/UrlConstants';
```
rushtong: unused

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+    super(props);
+    let params = new URLSearchParams(this.props.location.search);
+    this.state = {
+      consent: {
+        projectKey: '',
+        summary: ''
+      },
+      showError: false,
+      restriction: this.initRestriction(),
+      create: params.get('create') !== undefined ? true : false,
```
rushtong: Can just be `create: params.get('create') !== undefined,` 
I also just learned about [lodash isNil](https://lodash.com/docs/4.17.15#isNil)

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+      publicationResults: restriction !== undefined ? restriction.publicationResults : '',
+      genomicResults: restriction !== undefined ? restriction.genomicResults : '',
+      geographicalRestrictions: restriction !== undefined ? restriction.geographicalRestrictions : '',
+      other: restriction !== undefined ? restriction.other : '',
+      manualReview: restriction !== undefined ? restriction.manualReview : false,
+      comments: restriction !== undefined ? restriction.comments : '',
+      populationRestrictions: restriction !== undefined ? this.getAutocompleteData(restriction.populationRestrictions) : [],
+      genomicSummaryResults: restriction !== undefined && !isEmpty(restriction.genomicSummaryResults) ? restriction.genomicSummaryResults : '',
+      genomicPhenotypicData: restriction !== undefined ? restriction.genomicPhenotypicData : '',
+      consentPIName: restriction !== undefined && !isEmpty(restriction.consentPIName) ? restriction.consentPIName : '',
```
rushtong: This wall of `!== undefined` ternaries can be made more legible with some [lodash get](https://lodash.com/docs/4.17.15#get). For example this should work (untested), 
```
consentPIName: _.get(restriction, 'consentPIName', '')
```

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+    };
+    return resp;
+  }
+
+  init() {
+    spinnerService.show(RESTRICTION_SPINNER);
```
rushtong, line 94: Note for @LeoForconesi - this might impact your spinner PR.

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+        prev.restriction.diseaseRestrictions = [];
+        prev.restriction.generalUse = true;
+        return prev;
+      }, () => this.validateForm());
+    } else {
+      this.setFieldValue(field, value);
+    }
+  }
+
+  handleGenomicResultChange = (e, field, value) => {
```
rushtong: Looks unused

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+        prev.restriction.populationOriginsAncestry = true;
+        prev.restriction.hmbResearch = true;
+        return prev;
+      }, () => this.validateForm());
+    } else {
+      this.setFieldValue(field, value);
+    }
+  }
+
+  checkIfExist(data, key) {
```
rushtong: Looks unused

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -0,0 +1,864 @@
...
+            })
+          ]),
+          hr({ isRendered: this.state.restriction.genomicResults }, []),
+          div({ isRendered: this.state.restriction.genomicResults, style: styles.inputGroup }, [
+            p({ className: "inputFieldLabel" }, ["If ", i({}, ["Yes"]), ", please explain."]),
+            InputFieldTextArea({
+              disabled: false,
+              value: this.state.restriction.genomicSummaryResults,
+              name: 'genomicSummaryResults',
+              onChange: this.changeTextHandler
```
rushtong, line 791: Maybe here and for the input above you want `handleGenomicResultChange` ?

src/main/webapp/main/Routes.js
```diff
@@ -28,6 +30,9 @@ const Routes = ( props ) => (
     <Route path= {"/dataUseLetter/show"} render = {(routeProps) =>  <DataUseLetterIndex {...routeProps} {...props}/> }/>
     <Route path= {"/submissions/add-new"} render={(routeProps) => <SubmissionForm {...routeProps} {...props} />} />
     <Route path= {"/dataUse/show"} render = {(routeProps) =>  <DataUseRestrictionDetails {...routeProps} {...props}/> }/>
+
+    <Route path= {"/dataUse/restriction"} render = {(routeProps) =>  <DataUseRestrictionEdit {...routeProps} {...props}/> }/>
+
```
rushtong: nit: remove extra lines surrounding this route

src/main/webapp/util/UrlConstants.js
```diff
@@ -63,21 +64,24 @@ export const UrlConstants = {
...
   submissionSaveUrl: context + '/api/submissions/save-new',
   submissionDisplayUrl: context + '/api/submissions/display',
   submissionRemoveFileUrl: context + '/api/submission/remove-file',
-
   sampleConsentLinkUrl: context + '/api/sample-consent-link',
   sampleBreakLinkUrl: context + '/api/break-link',
   sampleApproveLinkUrl: context + '/api/approve-link',
   historyUrl: context + '/api/history',
+  submissionsUrl: context + '/api/submissions',
+  submissionDisplayUrl: context + '/api/submissions/display',
```
rushtong: Lines 80 and 81 are duplicates

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -22,14 +23,18 @@ class DataUseController extends AuthenticatedController {
...
+
+    def restriction() {
+      redirectToMainContainer()
     }
 
-    @Override
     def show() {
-        render(view: "/mainContainer/index", model:[restrictionId: params.id])
+      redirectToMainContainer()
     }
```
rushtong: Why do we need three methods to do the same exact thing? We should have a single method, `index` (or whatever, I don't really care) that does what `redirectToMainContainer` does. I suggest we revert the change in `AuthenticatedController`, delete these three, and then change whatever GSP or JS file tries to use these methods. We need to trim outdated code, not support it.

rushtong (changes requested):
See comments inline. Caveat, I did not test this page locally, just code reviewed. There is a lot of complicated logic in the original that I hope is preserved here, but this will need a review from the ORSP team for full functionality testing.

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
dags: done

grails-app/utils/org/broadinstitute/orsp/utils/IssueUtils.groovy
dags: done

src/main/webapp/dataUse/DataUseRestrictionEdit.js
dags: done

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
vvicario: I tried to do that but there is an issue. If we have different url in URLMappings with the same controller and action, it executes the first match, and we need different url to be handled by React router. If it executes the first url that it's found, then router  doesn't return the expected component. 
we will work on that in the following story:
https://broadinstitute.atlassian.net/browse/BTRX-637

rushtong: Why would you have:
> If we have different url in URLMappings with the same controller and action

? I'm not suggesting that you do that. In fact, I would argue that that is an error.

src/main/webapp/dataUse/DataUseRestrictionEdit.js
vvicario: Nice

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
LeoForconesi: Let me jump in here. 
Since Groovy is catching every route change before react (eg. when refreshing the browser's page), we needed a way to redirect these calls to our MainContainer view (where React router lives) using `redirectToMainContainer()`, which, in this case is defined in DataUseController. 
So in order to do this redirection, and match the url with the one present in react router we created methods that match with them when groovy gets these urls.
We used this only for pages resfresh, and return to previous page navigation.

rushtong: > Since Groovy is catching every route change before react (eg. when refreshing the browser's page)

Agreed. What I'm suggesting is that all of the various calls to the these methods be consolidated into a single API call. Those page calls can come from JS or GSP, it doesn't matter. It makes no sense to support three different code paths to the exact same behavior.

vvicario:
@rushtong can you re-review please?

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
vvicario: I renamed this method to save instead of create

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -89,6 +93,18 @@ class DataUseController extends AuthenticatedController {
         }
     }
 
+    def saveSdul() {
+        try {
+            DataUseRestriction restriction = dataUseLetterService.updateRestrictionFromParams(request.JSON)
```
rushtong: This method now throws an Illegal Argument exception, which should probably return a 400 bad request. 

From the previous round of feedback, it should also check for existence, which would return a 404 if not found.

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
```diff
@@ -1,6 +1,8 @@
 package org.broadinstitute.orsp
 
 import grails.gorm.transactions.Transactional
+import javassist.NotFoundException
```
rushtong: unused

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -89,6 +93,18 @@ class DataUseController extends AuthenticatedController {
...
+            dataUseLetterService.saveRestriction(restriction, getUser()?.displayName)
+            response.status = 200
+            render(response.status)
```
rushtong: In general REST practice, when an API saves an object, it should return the object just saved. That way callers can be sure the save really happened.

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
```diff
@@ -56,5 +58,13 @@ class DataUseLetterService {
         restriction
     }
 
+    DataUseRestriction updateRestrictionFromParams(Object params) {
+        if (params.consentGroupKey) {
+            DataUseRestriction existentRestriction = DataUseRestriction.findByConsentGroupKey(params.consentGroupKey)
```
rushtong, line 59: Should probably throw the not found exception here if not found.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -400,17 +400,21 @@ class QueryService implements Status {
...
+                    ' select * from storage_document ' +
+                            ' where consent_collection_link_id in :consentCollectionIds' +
+                            ' and deleted = 0'
+            final SQLQuery sqlQuery = session.createSQLQuery(query)
+            results = sqlQuery.with {
+                addEntity(StorageDocument)
+                setParameterList('consentCollectionIds', consentCollectionId)
+                list()
+            }
         }
```
rushtong: You don't need the redundant variable `results`. Just let the sqlQuery be the last line and then add an else with `Collections.emptyList()` so either the query results or the empty list is returned.

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
dags: unused. Please remove it.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
rushtong: Doesn't look like this feedback was addressed.

rushtong (changes requested):
See comments inline.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
vvicario: If consent group doesn't have a restriction it will be created in DataUseRestrictionParser. This is why I am not returning 404. If exists it uses the existent one. Does this clarify the logic?
dataUseRestriction = dataUseRestriction == null ? new DataUseRestriction() : dataUseRestriction

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
vvicario: If it is not found, it will be created in DataUseRestrictionParser

rushtong: Gotcha, yes, thank you!

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
rushtong: For the 404 question, yes, this makes sense that you don't need to handle not-found. The 400 handling is still not correct though, I'll comment again in the updated code.

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
```diff
@@ -89,6 +81,18 @@ class DataUseController extends AuthenticatedController {
         }
     }
 
+    def saveSdul() {
+        try {
+            DataUseRestriction restriction = dataUseLetterService.getRestrictionFromParams(request.JSON)
```
rushtong, line 86: Follow-up from previous feedback. This method can throw an illegal argument exception which is converted into a 500 below. We should catch that case and return a BadRequest (400) instead.

grails-app/services/org/broadinstitute/orsp/DataUseLetterService.groovy
```diff
@@ -42,9 +43,9 @@ class DataUseLetterService {
         }
     }
 
-    DataUseRestriction createSdul(DataUseRestriction restriction, String displayName) {
+    DataUseRestriction saveRestriction(DataUseRestriction restriction, String displayName) {
         Issue consent = queryService.findByKey(restriction.consentGroupKey)
-        def updatedOrCreated = "Created"
+        def updatedOrCreated = "Updated"
```
rushtong: This is previous code that I missed, sorry. Avoid `def` when possible. Also, this could be a lot simpler:
```
String updatedOrCreated = (restriction.id == null) ? "Created" : "Updated"
```

grails-app/utils/org/broadinstitute/orsp/utils/DataUseRestrictionParser.groovy
```diff
@@ -54,7 +54,7 @@ class DataUseRestrictionParser {
         dataUseRestriction.recontactMust = params.recontactMust
         dataUseRestriction.genomicPhenotypicData = params.genomicPhenotypicData
         dataUseRestriction.cloudStorage = params.cloudStorage
-        dataUseRestriction.irb = IssueUtils.getBooleanForParam(params.irb)
+        dataUseRestriction.irb =  IssueUtils.getBooleanForParam(params.irb)
```
rushtong: nit: extraneous spacing, this change can be reverted

src/main/webapp/PageNotFound/PageNotFound.js
```diff
@@ -0,0 +1,21 @@
...
+     return (
+       div({className:"row", style: {"margin":"60px 0"}}, [
+        div({className:"col-xs-12"}, [
+          h1({}, ["Sorry, this page could not be found."])
+         ])
+       ])
+     )
+  }
+})
+export default NotFound;
```
rushtong, line 21: This feedback doesn't necessarily need to be applied, but this component can be a [pure functional component](https://logrocket.com/blog/pure-functional-components/) instead. There's no props or state that are used in the render (although those can be used in functional components too...). Something like this should work and be a lot fewer lines of code:
```
export const NotFound = () => {
  return (
    div({className:"row", style: {"margin":"60px 0"}}, [
     div({className:"col-xs-12"}, [
       h1({}, ["Sorry, this page could not be found."])
      ])
    ])
  );
};

```

src/main/webapp/main/Routes.js
```diff
@@ -28,6 +30,8 @@ const Routes = ( props ) => (
     <Route path= {"/dataUseLetter/show"} render = {(routeProps) =>  <DataUseLetterIndex {...routeProps} {...props}/> }/>
     <Route path= {"/submissions/add-new"} render={(routeProps) => <SubmissionForm {...routeProps} {...props} />} />
     <Route path= {"/dataUse/show"} render = {(routeProps) =>  <DataUseRestrictionDetails {...routeProps} {...props}/> }/>
+    <Route path= {"/dataUse/restriction"} render = {(routeProps) =>  <DataUseRestrictionEdit {...routeProps} {...props}/> }/>
+    <Route path= {"/*"} render = {(routeProps) =>  <PageNotFound {...routeProps} {...props}/> }/>
```
rushtong, line 34: The not found component is a recent addition to this PR. Has `path= {"/*"}` been tested? I'm thinking it won't have any effect on the non-react pages, but we should be sure.

rushtong (changes requested):
One minor change requested which is to properly handle the 400 case. Also, the Not Found component is an unnecessary addition to this PR. It really should be in it's own ticket and not part of this one.

src/main/webapp/main/Routes.js
dags: This is intended to work in combination with urlMappings change. Anything not resolved in urlMappings is redirected to be handled by Routes. And if not found there a 404 page is shown. It was tested late yesterday and worked as expected, with a still not resolved issue of getting an additional 500 error in notFound page. This change is related to handle yor comment about "redirectToMainContainer" .

grails-app/controllers/org/broadinstitute/orsp/api/DataUseController.groovy
dags: additional catch added

src/main/webapp/main/Routes.js
rushtong: The conversation around `redirectToMainContainer` was related to having three different methods performing the same function in a Groovy controller method. I don't see the connection to this. In either case, it's a relatively small concern.

rushtong (approved):

dags (approved):

src/main/webapp/main/Routes.js
dags: Will try to explain it. We had 3 methods performing same function on Groovy controller because were handling 3 different urls (vies, pages). Since those pages are implemented in React now, but called from Grails by now, we need to go to the Grails controller and redirect to React Router which will handle the request at the end. While the 3 methods looked the same, redirectToMainContainer call was different on each case because the payload: the requested URL. That requested URL ended in Routes and was handled by Switch and Router. So, we removed actions from Grails controller and add a "generic" urlmapping in urlMappings to handle any not-matcher url and redirect it to React Router. Since any non existing url will end in react router as well, we added the /* route at end to catch all non exiting pages. That is the idea. Not sure if the best, at least the best we had. :-)

rushtong: I think this might be the disconnect ...

> Since those pages are implemented in React now, but called from Grails by now

If we have pages implemented in React, then there should be no calls from Grails to those pages. It sounds like there's a circular route handling process here that could use some de-convoluting. I wish I could spend some time simplifying this ... just not possible right now.

dags: I agree, there is something that needs improvement. We already have some stories in this Sprint for this, converting remaining pages/views to React.

PR #248, opened 2019-08-30
Handle About Text and User Session

wloforteBroad:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-771

## Changes
- Move about text to a component
- Handle user session and show info depending If the user is logged in or not

## Testing
Check About text logged in and after log out.

---


src/main/webapp/components/About.js
```diff
@@ -0,0 +1,74 @@
...
+      };
+  }
+
+  hasSession = () => {
+      User.isAuthenticated().then(resp => {
+        this.setState(prev => {
+          prev.logged = resp.data.session;
+          return prev;
+        })
+      }).catch(error => console.log(error));
```
rushtong: Should you use `ErrorHandler.js` here?

src/main/webapp/components/About.js
```diff
@@ -0,0 +1,74 @@
...
+            isRendered: this.state.logged && accessDetails,
+            href:"https://iwww.broadinstitute.org/sponsored-research/research-subject-protection/office-research-subject-protection"}, [
+              "ORSP on the Broad Intranet"
+          ]),
+        ]),
+        p({ style: { fontFamily : styles.fontFamily, fontSize: styles.textFontSize }}, [
+          "The Broad's Office of Research Subject Protection (ORSP) is committed to helping our community adhere to " +
+          "federal regulations and institutional policies governing the protection of human subjects who make our" +
+          "research possible. To fulfill that mission, the Broad must ensure appropriate regulatory oversight and" +
+          "reliable documentation storage.  The ORSP Portal is an online platform where Broad staff members can upload " +
```
rushtong, line 52: Note the spacing around quotes between lines. Does the `...and" + "reliable...` run together?

src/main/webapp/components/About.js
```diff
@@ -0,0 +1,74 @@
...
+            href:"mailto:orsp-portal@broadinstitute.org"}, ["orsp-portal@broadinstitute.org"]
+          ), " for assistance."
+        ]),
+        div({ isRendered: this.state.logged && accessDetails }, [
+            h3({ style: { fontSize: styles.titleSize }}, ["User Guide"]),
+            p({ style: { fontFamily : styles.fontFamily, fontSize: styles.textFontSize }},[
+              "To access detailed instructions about how to use the ORSP portal, please visit:",
+              br({}),
+              a({
+                href:"https://intranet.broadinstitute.org/research-subject-protection/orsp-online-portal-submission-system"},
```
rushtong: Just a thought ... should this open in a new tab? It probably doesn't now, but I think it might be a good idea to do that here.

rushtong (approved):
Looks good, some questions inline. I think the only thing I feel strongly about is the console log. If you want to address that, then 👍

src/main/webapp/components/About.js
wloforteBroad: It doesn't, but you're right it's indeed a good idea.

grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
```diff
@@ -322,6 +322,14 @@ class AuthenticatedController implements Interceptor, UserInfo {
...
     }
 
+    def hasSession() {
+        if (getUser()) {
+            render(['session': true] as JSON)
+        } else {
+            render(['session': false] as JSON)
+        }
+    }
+
```
dags, line 328: Assuming getUser() returns either true or false, may use  : 

```groovy
render(['session': getUser()] as JSON) 
```

src/main/webapp/components/About.js
```diff
@@ -0,0 +1,74 @@
...
+          logged: false,
+      };
+  }
+
+  hasSession = () => {
+      User.isAuthenticated().then(resp => {
+        this.setState(prev => {
+          prev.logged = resp.data.session;
+          return prev;
+        })
```
dags, line 25: No need for prev here : 
```javascript
this.setState({
     logged : resp.data.session
})
```

src/main/webapp/components/About.js
```diff
@@ -0,0 +1,74 @@
...
+      }).catch(error => this.setState(() => { throw error; }));
+  };
+
+  componentDidMount(){
+      this.hasSession();
+  }
+
+  render() {
+
+    const accessDetails = !isEmpty(this.props.showAccessDetails) ? true : false;
```
dags:
```javascript
const accessDetails = !isEmpty(this.props.showAccessDetails)
```
shouldn't be enough ?

grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
wloforteBroad: It returns session or null, but I'll use `render(['session': getUser()!=null] as JSON) `

dags (approved):

PR #250, opened 2019-09-03
[BTRX-757] Handle abort axios operation

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-757

## Changes
Added a flag to avoid updating the state of an unmounted component.
At least two more approaches have been tested and this one is the most efficient.
This approach was taken from here https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component
Also fixed the warnings messages in Data Use Restrictions Edit component.

## Testing

---


src/main/webapp/adminOnly/AdminOnly.js
```diff
@@ -48,43 +51,44 @@ export const AdminOnly = hh(class AdminOnly extends Component {
...
           prev.formData = formData;
           prev.initial = initial;
           prev.isAdmin = component.isAdmin;
           return prev;
         })
-      }).catch(error => {
-        this.setState(() => { throw error; });
-      });
+      }
+    }).catch(() => { });
```
dags: A catch-all block without a logger isn't a good idea. Please add logging here.

src/main/webapp/components/RequestClarificationDialog.js
```diff
@@ -41,8 +52,8 @@ export const RequestClarificationDialog = hh(class RequestClarificationDialog ex
   validateClarification() {
     let isValid = false;
     if (this.props.linkClarification === true) {
-      if (!isEmpty(this.state.clarification) && !isEmpty(this.state.pm) && 
-          !isEmpty(this.state.pm[0]) && !isEmpty(this.state.pm[0].key)) {
+      if (!isEmpty(this.state.clarification) && !isEmpty(this.state.pm) &&
```
dags, line 55: Please verify if we can use lodash's _.isEmpy here. Lodash is being imported where this isEmpty is defined anyway (src/main/webapp/util/Utils.js)

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -394,10 +406,10 @@ class NewConsentGroup extends Component {
   };
 
   handleInfoSecurityValidity(isValid) {
-    this.setState(prev => { 
+    this.setState(prev => {
```
dags: prev can be avoided here .

src/main/webapp/consentGroupContainer/ConsentGroupContainer.js
```diff
@@ -51,20 +58,24 @@ export const ConsentGroupContainer = hh(class ConsentGroupContainer extends Comp
...
-        return prev;
-      })
-    });
+      if (this._isMounted) {
+        this.setState(prev => {
+          prev.history = resp.data;
+          return prev;
+        });
+      }
+    }).catch(() => {});
```
dags, line 67: avoid empty catch-all. Throw error or log it.

src/main/webapp/consentGroupContainer/ConsentGroupContainer.js
```diff
@@ -51,20 +58,24 @@ export const ConsentGroupContainer = hh(class ConsentGroupContainer extends Comp
...
 
   // comments
   getComments() {
     Review.getComments(this.props.consentKey).then(result => {
-      this.setState(prev => {
-        prev.comments = result.data;
-        return prev;
-      })
+      if (this._isMounted) {
+        this.setState(prev => {
```
dags: prev can be avoided here ...

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -29,56 +31,64 @@ export const ConsentGroupDocuments = hh(class ConsentGroupDocuments extends Comp
...
-        this.setState(prev => {
-            prev.documents = JSON.parse(resp.data.documents);
-            prev.user = user.data;
-            return prev;
-          }, () => {
-            this.props.updateDocumentsStatus({ attachmentsApproved: resp.data.attachmentsApproved })
-          }
-        );
+        if (this._isMounted) {
+          this.setState(prev => {
```
dags, line 59: here too

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -29,56 +31,64 @@ export const ConsentGroupDocuments = hh(class ConsentGroupDocuments extends Comp
...
+            }, () => {
+              this.props.updateDocumentsStatus({ attachmentsApproved: resp.data.attachmentsApproved })
+            }
+          );
+        }
       });
-    }).catch(error => {
-      this.setState({ serverError: true });
-      console.error(error);
+    }).catch(() => {
```
dags, line 69: empty catch all  block, avoid it.

src/main/webapp/dataUse/DataUseRestrictionEdit.js
```diff
@@ -76,15 +81,15 @@ export const DataUseRestrictionEdit = hh(class DataUseRestrictionEdit extends Co
       pediatric: get(restriction, 'pediatricLimited', ''),
       collaborationInvestigators: get(restriction, 'collaborationInvestigators', ''),
       irb: get(restriction, 'irb', ''),
-      publicationResults: get(restriction, 'publicationResults', ''),
+      publicationResults: restriction !== undefined && !isEmpty(restriction.publicationResults) ? restriction.publicationResults : '',
```
dags: replace !== undefined with != null, here and all occurrences below

src/main/webapp/linkWizard/LinkWizard.js
```diff
@@ -74,18 +75,24 @@ export const LinkWizard = hh( class LinkWizard extends Component {
...
   getUserSession() {
-    User.getUserSession().then(
-      resp => this.setState({ user: resp.data })
-    ).catch(error => {
-      this.setState(() => { throw error; });
+    User.getUserSession().then( resp => {
+      if (this._isMount) {
+        this.setState({ user: resp.data })
+      }
+    }).catch(error => {
```
dags, line 92: Just a comment: we need to adopt a common criteria with catch and errors.

src/main/webapp/linkWizard/SelectSampleConsent.js
```diff
@@ -8,7 +8,7 @@ import { Panel } from '../components/Panel';
 import { AddDocumentDialog } from "../components/AddDocumentDialog";
 import { Table } from "../components/Table";
 import { CONSENT_DOCUMENTS } from '../util/DocumentType';
-import { ConsentGroup, SampleCollections } from "../util/ajax";
+import {ConsentGroup, SampleCollections} from "../util/ajax";
```
dags: add the removed spaces back.

src/main/webapp/components/RequestClarificationDialog.js
tadeork: I think it's OK to use this `isEmpty` and not `_.isEmpty`

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
tadeork: In this case is used the callback of the `setState`.

dags (approved):

src/main/webapp/adminOnly/AdminOnly.js
```diff
@@ -48,43 +51,46 @@ export const AdminOnly = hh(class AdminOnly extends Component {
...
           prev.initial = initial;
           prev.isAdmin = component.isAdmin;
           return prev;
         })
-      }).catch(error => {
-        this.setState(() => { throw error; });
-      });
+      }
+    }).catch(error => {
+      console.error(error);
```
vvicario, line 92: nit: remove this console error log

vvicario (approved):

src/main/webapp/components/RequestClarificationDialog.js
```diff
@@ -12,6 +12,9 @@ import { isEmpty } from '../util/Utils';
 import './ConfirmationDialog.css';
 
 export const RequestClarificationDialog = hh(class RequestClarificationDialog extends Component {
+
+  _isMounted = false;
```
rushtong, line 16: This is unused in this file.

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -71,10 +73,14 @@ class NewConsentGroup extends Component {
   }
 
   componentDidMount() {
+    this._isMounted = true;
     this.initDocuments();
     this.getUserSession();
-    ConsentGroup.getConsentGroupNames().then(
-      resp => this.setState({ existingGroupNames: resp.data }));
+    ConsentGroup.getConsentGroupNames().then(resp => {
+      if (this._isMounted) {
```
rushtong, line 80: I'm not sure why we are overriding how React handles component mounting. This line is always true since it is set to true on line 76. React will call this method right after the component mounts, so we should not have to manually track the lifecycle with `_isMounted`.

rushtong:
Will start a thread in slack to address the concerns here.

src/main/webapp/consentGroup/NewConsentGroup.js
tadeork: Actually it's not always true. Since this is a promise this will be send to the queue of the jvm so when the component unmounts the var `_isMounted` is setted to false and because is the callback of the promise it keeps the reference inside the promise. When the promise resolves it will have `_isMounted` properly set.

rushtong: Gotcha, yes, makes sense. Also addressed in slack.

PR #252, opened 2019-09-05
[BTRX-759] Convert landing page

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-759

## Changes
Added new page-component and redirects to intex from Route and Controllers. Removed about gsp page.

## Testing
Describe for the reviewer how to test the specifics of this PR, both positive and negative cases.

---

- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Get a thumb from Belatrix
- [ ] **Submitter**: Get a thumb from @rushtong
- [ ] **Submitter**: Merge to develop using github's "Squash and Merge" feature
- [ ] **Submitter**: Delete branch after merge

PR #216, opened 2019-07-22
BTRX - 616 - Convert Fundings Report

LeoForconesi:
## Addresses
[BTRX-616](https://broadinstitute.atlassian.net/browse/BTRX-616) : ORSP: Convert Funding Source Report

## Changes
- Implement a common bootstrap table 2 component
- Use this common component in Funding Report and Comments
- Remove Fundings report related grails files

## Testing
- From top nav options, navigate to Fundings Source Report. The table should be looking good. Funding reports is a paginated table, so pagination should also be working good.
- Export to excel should be working.
- Export to CSV should be working
- Print also should be working

---

- [x] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Get a thumb from Belatrix
- [x] **Submitter**: Get a thumb from @rushtong
- [ ] **Submitter**: Merge to develop using github's "Squash and Merge" feature
- [ ] **Submitter**: Delete branch after merge

PR #253, opened 2019-09-06
BTRX-760: ORSP: Convert "My Task List" AND "My Projects" page

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-761
https://broadinstitute.atlassian.net/browse/BTRX-760
## Changes
Add a new component IssueList to show my projects and my task list. Remove gsp views that are not been used
## Testing

---


src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,222 @@
+import { Component } from 'react';
+import { hh, div, h1, h } from 'react-hyperscript-helpers';
+import { TableComponent } from "../components/TableComponent";
+import { Project } from '../util/ajax';
+import { Spinner } from "../components/Spinner";
+import { spinnerService } from "../util/spinner-service";
+import { formatDataPrintableFormat } from "../util/TableUtil";
+import { printData } from "../util/Utils";
+import { Link } from 'react-router-dom';
+import { format } from 'date-fns';
```
tadeork: `format` is not used

src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,222 @@
...
+    hidden: true,
+    csvExport: false
+  }, {
+    dataField: 'projectKey',
+    text: 'Project',
+    sort: true,
+    headerStyle: (column, colIndex) => {
+      return { width: styles.project.projectKeyWidth };
+    },
+    formatter: (cell, row, rowIndex, colIndex) =>
```
tadeork, line 48: This links can be moved to a function like this 
https://github.com/broadinstitute/orsp-pub/pull/252/files#diff-a9f4747606a7ea1e1584bf172822bf8fR50
```suggestion
        formatter: (cell, row, rowIndex, colIndex) => {
    return div({}, [
      linkFormatter(row, row.title)
    ])
  }
```

src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,222 @@
...
+    this._isMounted = false;
+  }
+
+  init = () => {
+    spinnerService.show(SPINNER_NAME);
+    this.tableHandler(0, this.state.sizePerPage, this.state.search, this.state.sort, this.state.currentPage);
+  };
+
+  tableHandler = (offset, limit, search, sort, page) => {
+    if(this._isMounted) {
```
tadeork: `if (this._isMounted) { }` should be after the resolve of the promise is resolved

src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,222 @@
...
+  componentWillUnmount() {
+    this._isMounted = false;
+  }
+
+  init = () => {
+    spinnerService.show(SPINNER_NAME);
+    this.tableHandler(0, this.state.sizePerPage, this.state.search, this.state.sort, this.state.currentPage);
+  };
+
+  tableHandler = (offset, limit, search, sort, page) => {
```
tadeork: This params seems to be unused. I think it's possible to remove it and call directly `tableHandler` from `componentDidMount`

src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,222 @@
...
+          sizePerPageList: SIZE_PER_PAGE_LIST,
+          pagination: true
+        }),
+        h(Spinner, {
+          name: SPINNER_NAME, group: "orsp", loadingImage: component.loadingImage
+        })
+      ])
+    )
+  }
+})
```
tadeork: Missing semicolon.

src/main/webapp/rolesManagement/RolesManagement.js
```diff
@@ -1,5 +1,5 @@
 import { Component } from 'react';
-import { h, div, span } from 'react-hyperscript-helpers';
+import { h, div, span, h1 } from 'react-hyperscript-helpers';
```
tadeork: `span` is unused.

tadeork (commented):
One more comment about `RolesManagement` in the line 123 should be used `_isMounted` to prevent update unmounted component.

tadeork (approved):
:+1:

grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
```diff
@@ -0,0 +1,53 @@
...
+import org.broadinstitute.orsp.utils.IssueUtils
+import org.broadinstitute.orsp.utils.UtilityClass
+
+@Slf4j
+@Resource
+class IssueListController extends AuthenticatedController {
+
+    def getProjectsForUser() {
+        try {
+            new UtilityClass(queryService).registerIssueListMarshaller()
```
rushtong: I missed this in https://github.com/broadinstitute/orsp-pub/pull/216. We should never be writing new constructors like this `new UtilityClass(queryService)`. Instead, `queryService` should be a spring-injected bean like we do in all of our other classes. If a class needs access to a bean, then it is no longer a utility style class and is now something that should be inside the grails world, i.e. a domain class, a service class, a controller class, a job class, etc.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -150,6 +152,23 @@ class UtilityClass {
         }
     }
 
+    void registerIssueListMarshaller() {
+        JSON.createNamedConfig(ISSUE_LIST) {
+            SimpleDateFormat  sd = new SimpleDateFormat("yyyy-MM-dd")
+            it.registerObjectMarshaller( Issue ) { Issue ie ->
```
rushtong: Standard convention is to just use `it ->`. In this case, `ie` doesn't seem to be an abbreviation for anything recognizable.

grails-app/utils/org/broadinstitute/orsp/utils/UtilityClass.groovy
```diff
@@ -150,6 +152,23 @@ class UtilityClass {
...
+                return [
+                        id               : ie.id,
+                        projectKey       : ie.projectKey,
+                        summary          : IssueUtils.escapeQuote(ie.summary),
+                        status           : IssueUtils.escapeQuote(ie.status),
+                        type             : IssueUtils.escapeQuote(ie.type),
+                        updateDate       : ie.updateDate ? sd.format(ie.updateDate) : '',
+                        actors           : queryService.findUsersInUserNameList((List<String>)ie.getActorUsernames())?.collect { it.displayName }
```
rushtong: Why the need for casting here `(List<String>)ie.getActorUsernames()` ? Issue `getActorUsernames` already returns a `Collection<String>`

src/main/webapp/issueList/IssueList.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  init = () => {
+    spinnerService.show(SPINNER_NAME);
+    this.tableHandler();
+  };
+
+  tableHandler = () => {
+      Project.getProjectByUser(this.paramsContext.get('assignee'), this.paramsContext.get('max')).then(result => {
+        if(this._isMounted) {
```
rushtong, line 129: ~You probably would not need to check mounted state if this api call were instead moved to `componentDidMount()`~

On second thought, that might not work, `tableHandler` is called pretty often.

vvicario:
@rushtong Greg I worked on your feedback, can you please check it when you have some time?

rushtong (approved):
👍

PR #256, opened 2019-09-10
BTRX-737 ORSP - Tech debt - Create common exceptions handler

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-737
## Changes
Add Declarative Controller Exception Handling
## Testing

---


grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -63,8 +61,7 @@ class ClarificationController extends AuthenticatedController {
                 transitionService.handleIntake(issue, getProjectManagersForIssue(issue)*.userName)
                 response.status = 201
             } catch (Exception e) {
-                response.status = 500
-                render([error: "${e}"] as JSON)
```
rushtong, line 67: Now that JSON is not used here, `import grails.converters.JSON` unused. This is in several classes, so try running the changes through optimize imports to clean those up

grails-app/controllers/org/broadinstitute/orsp/api/ProjectController.groovy
```diff
@@ -114,8 +112,7 @@ class ProjectController extends AuthenticatedController {
             response.status = 200
             render([message: 'Project was deleted'] as JSON)
         } else {
-            response.status = 404
-            render([message: 'Project not found'] as JSON)
+            handleException(new NotFoundException('Project not found'), 404)
```
dags: not throwing an exception here ... please revert it.

grails-app/controllers/org/broadinstitute/orsp/api/ClarificationController.groovy
```diff
@@ -21,9 +21,7 @@ class ClarificationController extends AuthenticatedController {
         if (params.comment) {
             Comment comment = persistenceService.saveComment(issue.projectKey,  getUser()?.displayName, params.comment)
             if (comment == null || comment.hasErrors()) {
-                log.error("Error saving comment for issue '" + params.id + "': null")
-                response.status = 500
-                render([message: "Error saving comment"])
+              handleException(new Exception("Error saving comment"), 500)
```
dags: we are missing info here ?

grails-app/controllers/org/broadinstitute/orsp/api/ExceptionHandler.groovy
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.orsp.api
+
+import grails.converters.JSON
+import groovy.util.logging.Slf4j
+
+
+@Slf4j
+trait ExceptionHandler {
+
+    def handleException(Exception e, Integer statusCode) {
```
rushtong: I do think this is OK in this form, but it could be a bit smarter and more dev-friendly if this handles something similar to https://docs.oracle.com/javaee/7/api/javax/ws/rs/ClientErrorException.html and then you could get the status code from the exception itself. There could also be a handler for any exception case similar to this one. Pretty sure that library (javax.ws.rs-api) is not in this project - we have `org.springframework.web.client.HttpClientErrorException` which is much less helpful, ugh. 

Looking through the code, there are 3 cases, 400, 404, and 500. Might be easier to just have `handleNotFound`, `handleBadRequest`, and `handleException` that takes an optional string for the message.

rushtong (approved):
See comments inline, I don't feel too strongly here, so 👍 if you don't want to tackle specialized error handling.

tadeork (approved):

dags (approved):

PR #258, opened 2019-09-11
BTRX - 754 - HOC spinner

LeoForconesi:
## Addresses
[BTRX-754](https://broadinstitute.atlassian.net/browse/BTRX-754) : ORSP - Fetch Spinner to a HOC
## Changes
- Created a higher order component that holds the spinner
- Moved loading-image.svg to a react "assets" folder
- Pass showSpinner and hideSpinner through props where required
- Replace old spinner's methods.
- Remove unused code. 
- Moved New Project and Add Existing Consent Sample Data Cohort pages to react routes so they can have access to the spinner.
- Add missing spinners to: Project Review, Consent Group Review, and Landing Page. 

## Testing
- Every spinner in the app should be working as expected. Eitherway you can test as follows:
Access to a existing project, go to Comment's tab and add a new comment. The spinner should appear and disappear after the request is completed.
- There shouldn't be any other memory leak warning in console related to the spinner.

---


src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -1,6 +1,5 @@
 import { Component } from 'react';
-import { hh, h, p, div, h1, h2, h4, small, br, input, label, span, a, ul, li, button } from 'react-hyperscript-helpers';
-
+import { hh,p, div, h1, h2, h4, small, br, span, ul, li, button } from 'react-hyperscript-helpers';
```
dags: fix spacing ...

src/main/webapp/components/LoadingWrapper.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
dags: Let's handle hide and show in this component to decouple from App.js. App.js might be removed in that case.

LeoForconesi: :+1: I decided to keep App to construct a typical react structure so we can later add in here both header and footer here.

dags (approved):

src/main/webapp/main/App.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Routes from "./Routes";
+import LoadingWrapper from "../components/LoadingWrapper";
+
+const AppWithLoading = LoadingWrapper(Routes);
+
+export default function App() {
+
+    return (
+      AppWithLoading({
```
rushtong: Question ... do we need a constant for this? Can you just inline the LoadingWrapper in the return? Something like this:
```
return (LoadingWrapper(Routes));
```
I point this out because `AppWithLoading` is awkward. Tomorrow, if someone wants another app-level HOC, we then have something like 
```
const AppWithLoading = LoadingWrapper(Routes);
const AppWithMyNewFeature = MyNewFeature(AppWithLoading);
... etc.
```

src/main/webapp/components/LoadingWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+  let containerStyle = {
+    'position': 'fixed',
+    'top': '0',
+    'left': '0',
+    'width': '100%',
+    'height': '100%',
+    'background': 'rgba(255, 255, 255, 0.3)',
+    'zIndex': '9000'
+  };
+  let spinnerStyle = {'position': 'fixed', 'top': '30vh', 'left': '50vw', 'marginLeft': '-30px', 'zIndex': '10000'};
```
rushtong: Both of these can be constants

src/main/webapp/assets/loading-indicator.svg
```diff
@@ -0,0 +1 @@
+<svg width="60px"  height="60px"  xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" preserveAspectRatio="xMidYMid" class="lds-rolling" style="background: none;"><circle cx="50" cy="50" fill="none" ng-attr-stroke="{{config.color}}" ng-attr-stroke-width="{{config.width}}" ng-attr-r="{{config.radius}}" ng-attr-stroke-dasharray="{{config.dasharray}}" stroke="#2c3e50" stroke-width="9" r="40" stroke-dasharray="188.49555921538757 64.83185307179586" transform="rotate(324.641 50 50)"><animateTransform attributeName="transform" type="rotate" calcMode="linear" values="0 50 50;360 50 50" keyTimes="0;1" dur="1.2s" begin="0s" repeatCount="indefinite"></animateTransform></circle></svg>
```
rushtong: I see some ng-attributes here:

- ng-attr-stroke="{{config.color}}" 
- ng-attr-stroke-width="{{config.width}}" 
- ng-attr-r="{{config.radius}}" 
- ng-attr-stroke-dasharray="{{config.dasharray}}"

I'm guessing we copied this from an angular app somewhere. Pretty sure they're not necessary here.

src/main/webapp/components/Documents.js
```diff
@@ -156,6 +150,8 @@ export const Documents = hh(class Documents extends Component {
         isRendered: this.props.isConsentGroup === true
       }, [
           DataUseLetter({
+            showSpinner: this.props.showSpinner,
+            hideSpinner: this.props.hideSpinner,
```
rushtong: Just commenting here so I'm clear on the ownership. In both `AddDocumentDialog` and `DataUseLetter`, we are giving them the same spinner from the parent component. If true, is that intentional or should they have their own spinner?

rushtong: Additional question, if they are the same, will an action in `AddDocumentDialog` have any effect on the spinner behavior in `DataUseLetter`?

src/main/webapp/consentGroup/NewConsentGroup.js
```diff
@@ -127,13 +124,11 @@ class NewConsentGroup extends Component {
         this.state.user.displayName,
         this.state.user.userName)
         .then(resp => {
-          // TODO: window.location.href is a temporal way to redirect the user to project's consent-group page tab. We need to change this after
-          // transitioning from old gsps style is solved.
-          window.location.href = [component.serverURL, "project", "main?projectKey=" + component.projectKey + "&tab=consent-groups&new"].join("/");
-          spinnerService.hide(CONSENT_SPINNER);
+          this.props.history.push('/project/main?projectKey=' + qs.parse(this.props.location.search).projectKey + '&tab=consent-groups&new', {tab: 'consent-groups'});
```
rushtong, line 128: This is nice, but technically not related to the main thrust of the PR. In the future, we should strive to keep them separate. I'm fine with this for this PR though.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -155,6 +154,8 @@ export const ConsentGroupDocuments = hh(class ConsentGroupDocuments extends Comp
         actionLabel: 'Yes'
       }, []),
       Documents({
+        showSpinner: this.props.showSpinner,
+        hideSpinner: this.props.hideSpinner,
```
rushtong: Similar question here where we pass the parent component's spinner to the child.

src/main/webapp/main/LandingPage.js
```diff
@@ -66,19 +66,22 @@ const LandingPage = hh(class LandingPage extends Component{
...
+  componentDidMount = async () => {
     this._isMounted = true;
-    this.getProjectsList();
-    this.getTaskList();
-    this.hasSession();
-  }
+    this.props.showSpinner();
+    await this.getProjectsList();
+    await this.getTaskList();
+    await this.hasSession();
```
rushtong: Is there a way to await outside of componentDidMount? I say this because it looks like these three awaits turn them into serial calls instead of parallel ones.

src/main/webapp/main/LandingPage.js
```diff
@@ -98,11 +101,14 @@ const LandingPage = hh(class LandingPage extends Component{
           return resp;
         });
       }
-    }).catch(() => {});
+    } catch(e) {
+      this.props.hideSpinner()
```
rushtong: Does this have any impact on ErrorHandler? In other words, if there is an error, what takes precedence, this call or that one? Or is there no call to `ErrorHandler` here (which is what I think is true) ... and if so, should this component be wrapped?

rushtong: Additional thought ... should there be handing off of the component's spinner to the error handler?

src/main/webapp/main/LandingPage.js
```diff
@@ -117,25 +123,29 @@ const LandingPage = hh(class LandingPage extends Component{
...
+      const resp = await User.isAuthenticated();
+      if (this._isMounted) {
+        this.setState({
+          logged: resp.data.session
+        })
+      }
+    } catch(error) {
+      this.setState(() => {
+        throw error;
+      }, () => this.props.hideSpinner());
```
rushtong: Does it make more sense to hide the spinner before setting error state?

src/main/webapp/main/LandingPage.js
rushtong: Update: Everything (`App`) is wrapped in `ErrorHandler`, so my first question holds. Is there impact on spinner behavior when there is an error.

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -142,7 +137,7 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
   handleSuccessNotification = () => {
     const urlParams = new URLSearchParams(window.location.search);
     if (urlParams.has('new') && urlParams.get('tab') === 'consent-groups') {
-      history.pushState({}, null, window.location.href.split('&')[0]);
+      history.pushState({}, null, window.location.href.replace('&new',''));
```
rushtong: Is this a bug fix or related to the react upgrade?

src/main/webapp/components/Documents.js
dags: Current approach is to have a single spinner, wrapping top component. So, all components in Routes.js (Switch) will share same spinner. Traversing the component tree, we may or may not propagate same spinner. Or we could wrap an inner component with a new spinner. In this particular case, AddDocumentDialog and DataUseLetter will share same spinner (top spinner, the only one now) showing or hiding in different situations . New spinner component let us to wrap any component in it own spinner. We may need to review component tree and spinners, but I think it may be another story.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
dags: I think we need to create a story to add / re-arrange spinners, to be sure all components that handle async data loading are wrapped in a spinner.

src/main/webapp/projectContainer/ProjectContainer.js
```diff
@@ -113,6 +112,8 @@ export const ProjectContainer = hh(class ProjectContainer extends Component {
                 title: "Project Details",
               }, [
                   h(ProjectReview, {
+                    showSpinner: this.props.showSpinner,
+                    hideSpinner: this.props.hideSpinner,
```
rushtong: More of the same question here ... is it intentional that we pass the component's spinner down to these. I would imagine that it makes more sense to wrap ProjectReview with `LoadingWrapper`

src/main/webapp/main/LandingPage.js
dags: something like Promise.all or axios.all ? This patter is repeated in several components so we may find a general approach and fix all cases.

rushtong (commented):
Love it 👍👍👍👍 
Bunch of questions and some minor suggestions inline. I can re-review once those have some answers.

src/main/webapp/main/LandingPage.js
rushtong: Yes, exactly. Or, a separate async function that is called from componentDidMount and *that* function is `await`-ed from inside componentDidMount

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
rushtong: That description seems like *this* story to me. The passing in of the parent spinner is new code here. It would be just as easy to wrap `Documents` than it is to introduce an anti-pattern.

src/main/webapp/components/Documents.js
rushtong: I'm a little confused by this. The intention should have been to let every component have it's own spinner. Looking back at the ticket, I think the original intent was misunderstood. The idea behind introducing an HOC here was not to make the spinner global, but instead was to distribute the spinner handling down to each component that needed it. 

> Current approach is to have a single spinner, wrapping top component

☝️ That approach replicates what we had before, just a different implementation.

rushtong: Some links to previous comments that explain the problems with the global approach:
* https://github.com/broadinstitute/orsp-pub/pull/226#pullrequestreview-271412490
* https://github.com/broadinstitute/orsp-pub/pull/236#discussion_r313615963

src/main/webapp/main/LandingPage.js
LeoForconesi: We've just updated this, please check our approach, and let us know what do you think.

rushtong: Looks good!

grails-app/views/base/_topNav.gsp
```diff
@@ -25,7 +25,7 @@
                     <li class="dropdown">
                         <a href="#" class="dropdown-toggle" data-toggle="dropdown">New <b class="caret"></b></a>
                         <ul class="dropdown-menu">
-                            <li><a href="${createLink(controller: 'project', action: 'pages')}">New Project</a></li>
+                            <li><a href="${createLink(controller: 'project', action: 'newProject')}">New Project</a></li>
```
dags: this change is no longer related to this story. Do we really need it ?

src/main/webapp/components/InternationalCohorts.css
```diff
@@ -1,4 +1,4 @@
-.questionnaireProgressBar {
+.internationlCohortsBar .questionnaireProgressBar {
```
dags: this changes are no longer related to this story. Please remove it.

src/main/webapp/components/InternationalCohorts.js
```diff
@@ -86,7 +86,9 @@ export const InternationalCohorts = hh(class InternationalCohorts extends Compon
     }
 
     return (
-      QuestionnaireWorkflow({ questions: this.state.questions, handler: this.props.handler, determination: this.props.determination, questionnaireUnwrapped: true })
+      div({className: 'internationlCohortsBar'}, [
```
dags: this changes are no longer related to this story. Please revert it.

src/main/webapp/components/LoadingWrapper.js
```diff
@@ -0,0 +1,80 @@
...
+      } else if (overlaySpinner === 'fixedTop') {
+        style = {...baseStyle, ...containerStyleFixedTop }
+    } else {
+        style = {...baseStyle, ...containerStyle}
+      }
+      return style;
+    };
+
+    return (
+      h(Fragment, {},[
```
dags: this Fragment is useless, please remove it.

src/main/webapp/components/Comments.js
```diff
@@ -81,7 +82,9 @@ export const Comments = hh(class Comments extends Component {
   render() {
     return (
       h(Fragment, {}, [
-        TextEditor({
+        h(TextEditor, {
+          showSpinner: this.props.showSpinner,
```
dags: this props are no longer required, please remove it.

src/main/webapp/components/StatusBox.js
```diff
@@ -19,6 +24,10 @@ export const StatusBox = hh(class StatusBox extends Component {
       attachmentsApproved = ''
     } = this.props.status;
 
+    if (type !== '' && projectKey !== '') {
```
dags: looks like spinner makes no sense on this component. Please remove or justify its use.

dags: It is another good candidate for a functional component.

src/main/webapp/components/Wizard.css
```diff
@@ -61,7 +61,9 @@
 }
 
 .wizardStepTitle {
-  margin-bottom: 15px;
+  margin: 10px 0 15px 0;
+  font-size: 1.8rem;
```
dags: no longer needed here, please revert this change.

src/main/webapp/components/WizardStep.js
```diff
@@ -42,9 +42,11 @@ export const WizardStep = hh(class WizardStep extends Component {
     if (currentStep === step) {
       view = div({ className: "wizardStepContainer" }, [
         div({ className: "row" }, [
-          h2({ className: "wizardStepTitle " + (questionnaireStep !== undefined ? 'col-lg-8 col-md-7 col-sm-7 col-8' : 'col-lg-12 col-md-12 col-sm-12 col-12') }, [
-            small({}, ["Step " + (step + 1)]),
-            this.props.title
+          div({className: questionnaireStep !== undefined ? 'col-lg-8 col-md-7 col-sm-7 col-8' : 'col-lg-12 col-md-12 col-sm-12 col-12'}, [
```
dags: same here, revert it.

src/main/webapp/dataUse/DataUseRestrictionIndex.js
```diff
@@ -1,21 +1,22 @@
...
-import { div } from 'react-hyperscript-helpers';
+import React, { Component, Fragment } from 'react';
+import { h, hh } from 'react-hyperscript-helpers';
 import SampleCollectionLinks from './SampleCollectionLinks';
 import DataUseRestriction from './DataUseRestriction';
 
-class DataUseRestrictionIndex extends Component {
+const DataUseRestrictionIndex = hh(class DataUseRestrictionIndex extends Component {
 
   constructor(props) {
```
dags: constructor is useless, please remove it.

src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -10,15 +9,14 @@ import { InputFieldCheckbox } from '../components/InputFieldCheckbox';
...
-import { spinnerService } from "../util/spinner-service";
 import { MultiSelect } from "../components/MultiSelect";
 import { Search } from "../util/ajax";
 import { DataUse } from "../util/ajax";
 import _ from 'lodash';
 import { isEmpty } from "../util/Utils";
+import LoadingWrapper from "../components/LoadingWrapper";
 
-export const DataUseLetter = hh(class DataUseLetter extends Component {
+const DataUseLetter = hh(class DataUseLetter extends Component{
```
dags: revert space removal

src/main/webapp/profile/Profile.js
```diff
@@ -19,12 +19,12 @@ class Profile extends Component {
...
 
   init = () => {
-    spinnerService.showAll();
+    this.props.showSpinner();
     User.getUserSession(component.getUserUrl).then(resp => {
       this.setState(prev => { 
         prev.user = resp.data;
         return prev;
-      });
+      },() => this.props.hideSpinner());
```
dags: handle errors here

src/main/webapp/project/NewProjectDetermination.js
```diff
@@ -117,7 +117,7 @@ export const NewProjectDetermination = hh(class NewProjectDetermination extends
     return (
       WizardStep({ title: this.props.title, step: 1, currentStep: this.props.currentStep, questionnaireStep: true,
                    error: this.props.errors, errorMessage: 'Please answer the next question(s) above before moving to the next step'}, [
-        QuestionnaireWorkflow({ questions: this.state.questions, determination: this.props.determination, handler: this.props.handler })
+        QuestionnaireWorkflow({ questions: this.state.questions, determination: this.props.determination, handler: this.props.handler, questionnaire: 'determination' })
```
dags: revert this change since is not related to the story

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -247,7 +242,7 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
   };
 
   redirect = (action) => {
-    const path = action === 'new' ? UrlConstants.newConsentGroupUrl + '?projectKey='+ this.props.projectKey : UrlConstants.useExistingConsentGroupUrl;
+    const path = action === 'new' ? UrlConstants.newConsentGroupUrl + '?projectKey='+ this.props.projectKey : UrlConstants.useExistingConsentGroupUrl+ '?projectKey='+ this.props.projectKey;
```
dags, line 251: same as previous case commented by Greg. Keep both.

src/main/webapp/projectContainer/ConsentGroups.js
```diff
@@ -91,7 +87,6 @@ export const ConsentGroups = hh(class ConsentGroups extends Component {
 
   componentWillUnmount() {
     this._isMounted = false;
-    spinnerService._unregister(CONSENT_GROUPS_SPINNER);
   }
 
   getProjectConsentGroups = () => {
```
dags, line 94: must use new spinner, and handle errors

src/main/webapp/projectContainer/SubmissionForm.js
```diff
@@ -80,8 +79,7 @@ class SubmissionForm extends Component {
 
   componentWillUnmount() {
     this._isMounted = false;
```
dags, line 83: Please review if showSpinner is needed on componentDidMount and add it required.

src/main/webapp/projectDocument/ProjectDocument.js
```diff
@@ -5,7 +5,6 @@ import { PROJECT_DOCUMENTS } from '../util/DocumentType';
 import { ConfirmationDialog } from "../components/ConfirmationDialog";
 import { h, hh } from 'react-hyperscript-helpers';
 import { AlertMessage } from "../components/AlertMessage";
-import { Spinner } from '../components/Spinner';
 
 export const ProjectDocument = hh(class ProjectDocument extends Component {
```
dags: This component needs to be wrapped in spinner

src/main/webapp/main/App.js
LeoForconesi: This way to handle it is now deprecated, no need for any constants.

src/main/webapp/assets/loading-indicator.svg
LeoForconesi: Removed.

src/main/webapp/main/LandingPage.js
LeoForconesi: You're right, this has been modified.

src/main/webapp/projectContainer/ConsentGroups.js
LeoForconesi: It is a bug fix

src/main/webapp/projectContainer/ProjectContainer.js
LeoForconesi: Now each component is wrapped by our LoadingWrapper. No need to pass anything through props.

grails-app/views/base/_topNav.gsp
LeoForconesi: No we don't for this story, I've reverted all changes related to moving NewConsentGroup to React Routes.

src/main/webapp/components/StatusBox.js
LeoForconesi: Removed, I agree on making a functional component.

src/main/webapp/projectContainer/ConsentGroups.js
LeoForconesi: Done :+1:

src/main/webapp/projectContainer/SubmissionForm.js
LeoForconesi: It was needed, so I added our spinner wrapper.

src/main/webapp/main/LandingPage.js
LeoForconesi: As you suspected, we weren't handling any error here so ErrorHandler wouldn't have been triggered. I added its correspondent throw, hiding the spinner before that.

src/main/webapp/projectContainer/ProjectContainer.js
```diff
@@ -154,7 +153,8 @@ export const ProjectContainer = hh(class ProjectContainer extends Component {
                 key: "comments",
                 title: "Comments",
               }, [
-                  h(Fragment, {}, [Comments({
+                  h(Fragment, {}, [
```
dags: another not needed Fragment here

LeoForconesi: I've removed all these fragments, they all seem to be unnecessary.

LeoForconesi:
@rushtong this PR is ready to be re-reviewed.

src/main/webapp/adminOnly/AdminOnly.js
```diff
@@ -13,13 +13,10 @@ import { PREFERRED_IRB } from "../util/TypeDescription";
...
 import { Fundings } from "../components/Fundings";
-import { Spinner } from "../components/Spinner";
-import { spinnerService } from "../util/spinner-service";
 import { AlertMessage } from "../components/AlertMessage";
+import LoadingWrapper from "../components/LoadingWrapper";
 
-const ADMIN_ONLY_SPINNER = 'adminOnlySpinner';
-
-export const AdminOnly = hh(class AdminOnly extends Component {
+const AdminOnly = hh(class AdminOnly extends Component {
```
rushtong, line 18: Not a request for this PR, but I think an alternate syntax for this might look like:
```
export default LoadingWrapper(hh(class AdminOnly extends Component {... blah ... }), 'fixedTop') as AdminOnly
```
Might be worth looking at whether that saves boilerplate or makes it worse ... not really sure at this point.

src/main/webapp/components/LoadingWrapper.js
```diff
@@ -0,0 +1,78 @@
...
+
+    const hideSpinner = () => {
+      setLoading(false);
+    };
+
+    const getStyle = () => {
+      let style = {};
+      if (overlaySpinner === 'fixed') {
+        style = {...baseStyle, ...containerStyleFixed }
+      } else if (overlaySpinner === 'fixedTop') {
```
rushtong: Would be nice if these were constants defined somewhere. I can easily see bugs cropping up where this series of if's is expanded and devs mistype something. Also, might be better structured as a switch block.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
```diff
@@ -6,9 +6,9 @@ import { ConfirmationDialog } from "../components/ConfirmationDialog";
 import { h, hh } from 'react-hyperscript-helpers';
 import '../index.css';
 import { AlertMessage } from "../components/AlertMessage";
-import { Spinner } from '../components/Spinner';
-
-export const ConsentGroupDocuments = hh(class ConsentGroupDocuments extends Component {
+import LoadingWrapper from "../components/LoadingWrapper";
```
rushtong: Run these changes through the eslint rules for the project - you should see double quotes come up as warnings and should be single quotes. For example, when I optimize imports for this file, this is what it does for me:
```
import { Component, Fragment } from 'react'
import { h, hh } from 'react-hyperscript-helpers'
import { AlertMessage } from '../components/AlertMessage'
import { ConfirmationDialog } from '../components/ConfirmationDialog'
import { Documents } from '../components/Documents'
import LoadingWrapper from '../components/LoadingWrapper'
import '../index.css'
import { ConsentGroup, DocumentHandler, User } from '../util/ajax'
import { CONSENT_DOCUMENTS } from '../util/DocumentType'
```

src/main/webapp/dataUseLetter/DataUseLetter.js
rushtong: Unaddressed.

src/main/webapp/projectContainer/ConsentGroups.js
rushtong: Might be worth pulling out the bug fixes into a separate PR so they're not held up by this PR.

rushtong (approved):
Looks pretty good. Not sure if it makes sense to pull out the small bug fixes into a separate PR at this point - I'm fine either way. In general, I'm much more comfortable approving a tiny bug fix PR than I am with adding it as part of a feature. If a  feature PR needs to be reverted for some reason, the bug fixes embedded in it also get reverted, which is bad.

dags (approved):

src/main/webapp/adminOnly/AdminOnly.js
LeoForconesi: I discussed this with the team, we think it's more legible the way it is now. :smiley:

src/main/webapp/components/LoadingWrapper.js
LeoForconesi: It's true, eitherway we discussed styles with @jaime-bernales-broad and @wloforteBroad and we decided to remove one of these styles for a better user experience. The result is true or false condition now, so these constants are not needed anymore.

src/main/webapp/consentGroupDocuments/ConsentGroupDocuments.js
LeoForconesi: I will incorporate this practice into my code. Right now i've optimize all imports of the files that i've worked on this PR.

src/main/webapp/dataUseLetter/DataUseLetter.js
LeoForconesi: I've must have missed this. this space removal has been reverted now.

rushtong:
Still 👍

tadeork (approved):
:sunglasses: Great work!

PR #268, opened 2019-09-18
[BTRX-785] Download file as preview

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-785

## Changes
It's now possible to download a file as a way to preview it when it's selected to be uploaded.

## Testing

---


src/main/webapp/components/Table.js
```diff
@@ -92,15 +92,17 @@ export const Table = hh(class Table extends Component {
...
+    if (isEmpty(row.uuid)) {
+      urlObject = downloadSelectedFile(row.file);
     }
+
+    return a({
+      href: isEmpty(row.uuid) ? urlObject :`${UrlConstants.downloadDocumentUrl}?uuid=${row.uuid}`,
+      target: '_blank',
+      title: row.fileName,
+      download: row.fileName,
+    }, [row.fileName])
```
rushtong, line 105: Not a requirement for this PR, but the anchor tag + `window.URL.createObjectURL` approach did not work in FF. See https://github.com/DataBiosphere/duos-ui/pull/204 for how I fixed that in DUOS. Have not tested this locally (we don't have instructions for standing up a local instance against GCP yet, see  https://broadinstitute.atlassian.net/browse/BTRX-768) so this may work in FF - I recommend giving it a try. Pretty sure Chrome is our primary supported browser, so this may not be a high priority issue.

rushtong (approved):
👍 Looks good, pending the question of whether this needs to support FF.

src/main/webapp/components/Table.js
tadeork: I think it's a good idea. But I found a couple of issues: tried to implement it and couldn't armonice `js-file-download`with our code because the library triggers the click event without our control.

rushtong: 👍 Makes sense - I'm not necessarily committed to that specific implementation in ORSP. Please confirm that if this solution doesn't work in FF, then that is OK with Andrea's team. If it isn't OK, then you might need a different implementation.

tadeork: Looking at the code of the library I think I can implement just the part of that code which will work with our approach.

rushtong: Does that mean that the current approach does not work in FF? If so, is that a requirement that Andrea/Shiv need to have in place? It would be nice if I could run a branch locally, but we don't have a defined process for that yet.

wloforteBroad: @rushtong There are several things that are not working in FF. We haven't been testing FF, it was never a requisite to do it If I'm not mistaken but we can do it and fix the issues If needed of course. Having said that, this new feature that Tadeo added is working on FF, I tested it myself.

rushtong: Thank you @wloforteBroad - that is the answer the question I was asking. If it's working in FF, that's great - I'll look at this implementation again to see if makes more sense than the one I'm using in DUOS.

wloforteBroad: Great! We will merge this to feature then. Thanks Greg!

vvicario (approved):

PR #270, opened 2019-09-20
BTRX - 783 - NPE on Sign Out

LeoForconesi:
## Addresses
[BTRX-783](https://broadinstitute.atlassian.net/browse/BTRX-783): NPE on Sign Out

## Changes
- `issueIsForbidden` method in `AuthenticatedController` threw a Groovy handled error to the Ui since this is trying to access userName of a null object, returned by `getUser()` called in the parameters of `permissionService.issueIsForbidden`.

## Testing
- Log in with valid credentials
- Go to a Project
- Sign Out, There shouldn't be any error message on the ui
---

### DIFF
--- a/grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
+++ b/grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
@@ -328,7 +328,7 @@ class AuthenticatedController implements Interceptor, UserInfo, ExceptionHandler
     }
 
     def issueIsForbidden(issue) {
-        getUser() ? permissionService.issueIsForbidden(issue, getUser().userName, isAdmin(), isViewer()) : null
+        getUser() ? permissionService.issueIsForbidden(issue, getUser().userName, isAdmin(), isViewer()) : false
     }
 
     def redirectToMainContainer() {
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/broadinstitute/orsp/AuthenticatedController.groovy
```diff
@@ -328,7 +328,7 @@ class AuthenticatedController implements Interceptor, UserInfo, ExceptionHandler
     }
 
     def issueIsForbidden(issue) {
-        return permissionService.issueIsForbidden(issue, getUser().userName, isAdmin(), isViewer())
+        getUser() ? permissionService.issueIsForbidden(issue, getUser().userName, isAdmin(), isViewer()) : null
```
dags: looks like issueIsForbidden  is a boolean function, it is OK to return a null ?

LeoForconesi: You're right, I was trying to maintain the value it was returning before, but after a deeper look it is as you say, a boolean function. I've changed the return value to `false`.

dags (approved):

rushtong (approved):
👍

PR #274, opened 2019-09-20
Convert Top Nav Bar

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-622
## Changes
Create Top Nav Bar and Handle authentication router.
## Testing

---


src/main/webapp/components/TopNavigationMenu.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.navbar .inputField .inputFieldSelect .select__option.select__option--is-disabled {
+    background-color: #FAFAFA;
+    padding: 15px !important;
+}
+
+.navbar .inputField .inputFieldSelect .select__menu .select__menu-list .select__option {
+    padding: 20px 15px;
+}
```
rushtong, line 46: Preference is to use inline styling as much as possible. Most of the content this file can be replaced with style objects in the components where they are used. I think the one thing that can't is 
```
.navbar .inputField .inputFieldSelect .select__option.select__option--is-focused:hover {
    color: #337ab7;
}
```
but that could be a single class (i.e. `navHover` or whatever makes the most sense in this context) that is added to the className field for whatever component needs it.

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,258 @@
...
+      }
+    }
+  }
+
+  onSuccess = async(token) => {
+    await User.signIn(token);
+    Storage.setUserIsLogged(true);
+    this.init().then(resp => {
+      if (Storage.getLocationFrom() != null) {
+        this.props.history.push(Storage.getLocationFrom());
```
rushtong, line 122: TOL (thinking out loud): This could lead to a 404 if the from url is not properly formed. Wondering if we could leverage our `Routes.js` somehow to check whether it's worth following or instead, showing a message to the user that hey, we can't get you to `/restricted/passwords`, but here's the list of projects instead.

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,258 @@
...
+          } else if (item.linkDisabled === true) {
+            label =
+              span({ style: styles.listResultContainer }, [
+                span({ style: styles.badgeContactAccess }, [
+                  ' Please contact ' + item.reporter + ' for access'
+                ]),
+                div({ style: { 'marginTop': '10px' } }, [
+                  item.label
+                ])
+              ])
```
rushtong: This would read better if the label generation were a function so that so you can avoid the awkward if/else-if construct here. In the return block below, it should just call that function with the right data, i.e. something like 
```
          return {
            key: item.id,
            value: item.value,
            label: this.generateLabel(item),
            linkDisabled: item.linkDisabled,
            url: item.url,
            isDisabled: item.linkDisabled
          };
```
And in that label generation function, the only thing that looks different is item pm vs item reporter. Could be simplified with a ternary on that line instead of the if/else-if altogether.

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  isAdmin() {
+    return this.state.isLogged && this.state.userSession != null &&
+      (this.state.userSession.isAdmin || this.state.userSession.isORSP);
+  }
+
+  isViewer() {
+    return this.state.isLogged && this.state.userSession != null && this.state.userSession.isViewer;
+  }
```
rushtong: nit: line break between functions

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,258 @@
...
+              span({ className: "sr-only" }, ["Toggle navigation"]),
+              span({ className: "icon-bar" }, []),
+              span({ className: "icon-bar" }, []),
+              span({ className: "icon-bar" }, [])
+            ]),
+            h(Link, { className: "navbar-brand", to: { pathname: UrlConstants.index } },
+              [
+                span({}, [
+                  "ORSP Portal ",
+                  span({isRendered: process.env.NODE_ENV === 'development', className: "label label-danger" }, ["Dev"])
```
rushtong: Note that we also have a staging environment. I don't know how our node env is being set for either that or for dev ... will be looking for that now that we're actually using it.

rushtong: See https://github.com/broadinstitute/orsp-pub/blob/feature/.circleci/config.yml 
I'm not sure that `process.env.NODE_ENV` is actually set anywhere ... please confirm that it will be, or how this is set as I don't see it in this PR.

src/main/webapp/components/WizardStep.js
```diff
@@ -42,9 +42,11 @@ export const WizardStep = hh(class WizardStep extends Component {
...
       view = div({ className: "wizardStepContainer" }, [
         div({ className: "row" }, [
-          h2({ className: "wizardStepTitle " + (questionnaireStep !== undefined ? 'col-lg-8 col-md-7 col-sm-7 col-8' : 'col-lg-12 col-md-12 col-sm-12 col-12') }, [
-            small({}, ["Step " + (step + 1)]),
-            this.props.title
+          div({className: questionnaireStep !== undefined ? 'col-lg-8 col-md-7 col-sm-7 col-8' : 'col-lg-12 col-md-12 col-sm-12 col-12'}, [
+            h2({ className: "wizardStepTitle" }, [
+              small({}, ["Step " + (step + 1)]),
+              this.props.title
+            ])
```
rushtong, line 49: nit: This looks like an unrelated tweak to the UI. Fine for this PR, but in the future, we should strive to keep PRs as small as possible by submitting bug fix PRs separately from feature PRs.

grails-app/views/layouts/main.gsp
```diff
@@ -16,69 +16,13 @@
...
-    @import "${request.contextPath}/assets/media/css/TableTools.css";
-    .ui-autocomplete-loading {
-        background: white url('${request.contextPath}/assets/spinner.gif') right center no-repeat;
-    }
-    @media (min-width: 765px) and (max-width: 1200px) {
-        body {
-            margin-top: 110px;
-        }
-    }
     </style>
```
rushtong: Invalid closing tag

grails-app/views/layouts/main.gsp
```diff
@@ -16,69 +16,13 @@
...
-    <script src="https://apis.google.com/js/platform.js" async defer></script>
-    <meta name="google-signin-client_id" content="${grailsApplication.config.googleSignInClientId}">
     <script>
         if (location.protocol !== "https:") location.protocol = "https:";
     </script>
```
rushtong, line 21: Not your code here, but I was hoping we would not need this once the sign in page was converted over.

src/main/webapp/components/InternationalCohorts.css
```diff
@@ -9,17 +9,19 @@
     box-shadow:         0 1px 3px 1px #E1E1E1;
 }
 
-.questionContainer {
+.internationlCohortsBar .questionContainer {
     width: 323px;
     margin: 0 0 0 15px;
     padding-top: 20px;
 }
```
rushtong, line 16: Any time we touch a css file, we should be trying to extract code out of it instead of adding to it. This is a good example where a simple inline style object would suffice.

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,260 @@
...
+
+  isBroadUser() {
+    return this.state.isLogged && this.state.userSession != null &&
+      this.state.userSession.isBroad;
+  }
+
+  isAdmin() {
+    return this.state.isLogged && this.state.userSession != null &&
+      (this.state.userSession.isAdmin || this.state.userSession.isORSP);
+  }
```
rushtong: This seems like an alternate pattern for determining if the user is an admin or not. Why not use `component.isAdmin` and make any necessary changes there so they are consistent through out the application? Same thing applies to `isViewer` and `isBroadUser`.

src/main/webapp/consentGroupReview/ConsentGroupReview.js
```diff
@@ -202,7 +203,7 @@ const ConsentGroupReview = hh(class ConsentGroupReview extends Component {
                 prev.current = current;
                 prev.future = future;
                 prev.futureCopy = futureCopy;
-                prev.isAdmin = component.isAdmin;
+                prev.isAdmin = Storage.getCurrentUser() != null ? Storage.getCurrentUser().isAdmin : false;
```
rushtong: This extra boilerplate, if carried through, would result in a ton of changes throughout the app (just search on `component.isAdmin`). This also looks unrelated to the nav bar conversion. I suggest these changes be reverted and then re-implemented throughout the app in a way that makes sense, doesn't require so much boilerplate, etc.

src/main/webapp/dataUseLetter/DataUseLetter.js
```diff
@@ -381,7 +381,8 @@ const DataUseLetter = hh(class DataUseLetter extends Component {
       DUL.updateDUL(form).then(resp => {
         this.createRestriction();
         DUL.createDulPdf({ uid: id }).then(() => {
-          window.location.href = component.serverURL + "/dataUseLetter/show?id=" + id;
+          this.props.init(id);
+          this.props.hideSpinner();
```
rushtong: This looks unrelated to the top nav conversion. Please revert and submit as a bug fix with appropriate ticket.

src/main/webapp/dataUseLetter/DataUseLetterIndex.js
```diff
@@ -16,7 +16,8 @@ class DataUseLetterIndex extends Component {
     this.state = {
       isLoading: true,
       error: '',
-      dul: {}
+      uuid: '',
+      dul: {uuid: null}
```
rushtong: TOL: Not for this PR, but one of the areas I've had difficulty with in our current coding practice is the instantiation of state objects this way. Without the parameters defined as you have here, you see undefined errors, so this makes sense. However, this is not sustainable in the long run because every component that needs to work with a `dul` now needs to repeat this same boilerplate over and over again.

What we should be doing here instead is referring to a standardized JS model object elsewhere in the app that defines what a `dul` really is. That way any component get a state dul by calling the function. 

```
function dul() {
    return { uuid: null }
}
```

src/main/webapp/main/AuthenticatedRoute.js
```diff
@@ -0,0 +1,27 @@
...
+
+  const { path, location } = rest;
+
+  return (
+    <Route
+      path={path}
+      location={location}
+      render={
+        props =>
+       verifyUser(admin, props)
```
rushtong: nit: spacing here, and also throughout the file. In intelli-j, `command + option + L` will reformat the code with the correct eslint settings defined for the project.

src/main/webapp/main/LandingPage.js
```diff
@@ -119,15 +124,19 @@ const LandingPage = hh(class LandingPage extends Component{
...
-      })
+      });
+      Storage.setUserIsLogged(user.data.session);
     }
   };
 
   render() {
+    if (Storage.userIsLogged() && !this.state.logged) {
+      this.init();
+    }
```
rushtong: This is already in `componentDidMount` ... do we need to call it a second time before rendering?

src/main/webapp/main/LandingPage.js
```diff
@@ -119,15 +124,19 @@ const LandingPage = hh(class LandingPage extends Component{
...
     return (
       div({}, [
         About(),
-        div({className: "row", isRendered: this.state.logged && component.isBroad}, [
+        div({className: "row", isRendered: Storage.userIsLogged() &&  !nonBroadUser()}, [
```
rushtong: Instead of the negation `!nonBroadUser()`, can we use `isBroad` here? It would also make sense that `isBroad` would imply that you are also logged in, so these should, in theory, be combined into a single method call.

src/main/webapp/projectReview/ProjectReview.js
```diff
@@ -202,7 +203,7 @@ const ProjectReview = hh(class ProjectReview extends Component {
                   prev.futureCopy = futureCopy;
                   prev.editedForm = JSON.parse(data.data.suggestions);
                   prev.reviewSuggestion = true;
-                  prev.isAdmin = component.isAdmin;
+                  prev.isAdmin = Storage.getCurrentUser() != null ? Storage.getCurrentUser().isAdmin : false;
```
rushtong: See comments in `ConsentGroupReview.js` on this pattern.

vvicario:
@rushtong This PR is ready for review. If you think we need to split something else, please let me know! Thanks!

src/main/webapp/util/Storage.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+
+  getLocationFrom: () => {
+    return localStorage.getItem(URLFrom) ? JSON.parse(localStorage.getItem(URLFrom)) : null;
+  },
+
+  removeLocationFrom: () => {
+    return localStorage.removeItem(URLFrom);
+  }
+};
```
rushtong, line 38: Perhaps not for this PR, but I have some reservations about using local storage like this. I've found it to be problematic in duos. [Here's how they are doing this in Terra](https://github.com/DataBiosphere/terra-ui/blob/58d645ae53b08c7be255f199d09394f5d152a472/src/libs/browser-storage.js). I will do some research on alternatives, but for now, I'm not requesting any changes here.

src/main/webapp/components/Documents.js
```diff
@@ -99,6 +99,7 @@ export const Documents = hh(class Documents extends Component {
   };
 
   render() {
+    console.log('is viewer', component.isViewer)
```
rushtong: nit: revert

src/main/webapp/components/Footer.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react'
-import ReactDOM from 'react-dom'
-import { a, div } from 'react-hyperscript-helpers'
+import { Component } from 'react'
+import { a, div, hh, footer } from 'react-hyperscript-helpers'
+import { styles } from '../util/ReportConstants';
 
-export class Footer extends React.Component {
+export const Footer = hh(class Footer extends Component {
```
rushtong: nit: This doesn't need to be a full class - it can be a simple function or a functional component.

grails-app/views/layouts/main_component.gsp
```diff
@@ -42,12 +42,11 @@
         }
     }
     </style>
-    <script src="https://apis.google.com/js/platform.js" async defer></script>
+    <script src="https://apis.google.com/js/platform.js"></script>
```
rushtong: IIRC, async defer is the recommended approach: https://developers.google.com/+/web/api/javascript

grails-app/views/layouts/main_component.gsp
```diff
@@ -85,7 +84,8 @@
         contextPath: "${request.contextPath}",
         isAdmin: ${session.isAdmin ? session.isAdmin : false},
         isViewer: ${session.isViewer ? session.isViewer : false},
-        isBroad: ${session.isBroad ? session.isBroad : false}
+        clientId: "${grailsApplication.config.googleSignInClientId}"
+
       };
     </script>
```
rushtong, line 93: For a separate cleanup story, there are several functions here that look unused post-conversion:

    loadComments
    loadHistory
    toggleContinueMessage
    togglePropertyMessage

rushtong (commented):
See feedback inline. Since the latest merge from develop, some of my comments might not make sense as I've been writing them since early this morning. Will re-review when I get another chance.

src/main/webapp/dataUseLetter/DataUseLetterIndex.js
```diff
@@ -16,7 +16,8 @@ class DataUseLetterIndex extends Component {
     this.state = {
       isLoading: true,
       error: '',
-      dul: {}
+      uuid: '',
```
LeoForconesi: This seems to be unused.

src/main/webapp/main/Routes.js
```diff
@@ -17,33 +17,36 @@ import DataUseRestrictionEdit from '../dataUse/DataUseRestrictionEdit';
...
+    <AuthenticatedRoute path= {"/dataUse/view"}  component={DataUseRestrictionDetails} props={props}/> }/>
+    <AuthenticatedRoute path= {"/consent-group/use-existing"} component={LinkWizard} props={props}/> }/>
+    <AuthenticatedRoute path= {"/consent-group/new"} component={NewConsentGroup} props={props}/> }/>
+    <AuthenticatedRoute path= {"/dataUse/restriction"} component={DataUseRestrictionEdit} props={props}/> }/>
+    <AuthenticatedRoute path= {"/issueList/list"} component={IssueList} props={props} /> }/>
+    <AuthenticatedRoute path= {"/search/index"} component={Search} props={props} /> }/>
+    <AuthenticatedRoute path= {"/submissions/add-new"} component={SubmissionForm} props={props} /> }/>
+    <Route path= {"/infoLink/showInfoLink"} component={InfoLink} props={props}/> }/>
+    <Route path= {["/index", "/"]} exact component={LandingPage} props={props}/> }/>
+    <Route path= {["/index/profile", "/profile"]} render = {(routeProps) =>  <Profile {...routeProps} {...props}/> }/>
```
LeoForconesi: `/infoLink/showInfoLink` and `/index/profile` should be wrapped in AuthenthicatedRoute.

src/main/webapp/main/LandingPage.js
vvicario: Greg, When user is not logged, the component is been created with "logged = false". After user logs in, we display the same page and "logged" is still false.  We need some way to change the state of this component.

src/main/webapp/components/TopNavigationMenu.css
jaime-bernales-broad: The problem I'm having with inline styling are class names (such as: ".select__menu-list") and how they are created/rendered, they are auto-generated components and we aren't assigning them className or style directly. Is for that reason I'm editing appearance via CSS, because I'm not getting access to them via inline. Also the chain of concatenated classes is a problem to me for reproducing it inline. Any help is welcome

grails-app/views/layouts/main_component.gsp
vvicario: US: https://broadinstitute.atlassian.net/browse/BTRX-793

src/main/webapp/components/TopNavigationMenu.css
rushtong: We *should* be assigning the styles directly, that's the point of having inline styles. First, you create the style object, then you use that object as a component property, i.e. `ComponentX({style: {property: "value"}});` or `ComponentX({style: styleObject.myComponent});`

Shared CSS requires far more bug checking than inline styles. If I make a change to a classname css, then I need to physically view all usages of that css to ensure no bugs were introduced. Isolating the CSS to the component that uses it reduces the number of bugs introduced.

src/main/webapp/main/LandingPage.js
rushtong: The main problem is that we're doing this in the render. Render is called every time state changes. The init method changes state, so you get into an infinite loop situation. You need to handle the state changing events in the appropriate lifecycle handlers.

src/main/webapp/search/Search.js
```diff
@@ -6,6 +6,7 @@ import ProjectAutocomplete from "../util/ProjectAutocomplete";
 import SearchResults from "./SearchResults";
 import UserAutocomplete from "../util/UserAutocomplete";
 import { UrlConstants } from "../util/UrlConstants";
+import "./style.css";
```
rushtong, line 9: @jaime-bernales-broad This is one example of where shared css is a maintenance headache. This PR is about converting the top nav bar to react. However, this addition means that I, as a reviewer, now need to look at the search page to ensure that there are no UI bugs introduced by the addition of this css file. It's unlikely that there is, but still, it complicates both the development and review processes with extra time spent reviewing unrelated code. 

The other example is in `style.css` itself. Due to the number of deletions in that file, both the developer and the reviewer need to look at it. I can find usages of styles in that file in quite a few places so each of them would need to be reviewed to ensure no bugs were introduced.

rushtong:
@vvicario I found at least one file where there is a reference to a deleted file. Look at `unguarded_component.gsp` - that references `_topNav.gsp` (and others) which is now deleted. There may be others, please re-scan for the deleted files to make sure we haven't missed any other references.

src/main/webapp/components/TopNavigationMenu.js
vvicario: Hi Greg, I'll take a look, I've seen it in UrilConstants,js
export const context = process.env.NODE_ENV === 'test' ? 'http://localhost' : component.contextPath;

src/main/webapp/search/Search.js
vvicario: HI Greg I added this css because we are not using any more index.js to open the page. We redirect from router, I removed search/index.js where we had  import  style.css, if we open search component from Router, without it, search gets visually broken. Now we are impoting the css file directly in the search.js. We know about the reviewing work on this but it was a neccesary change to use the router

src/main/webapp/components/TopNavigationMenu.js
rushtong: The only other place where I think that var might have existed is in the Dockerfile, but I don't see it there either. https://github.com/broadinstitute/orsp-pub/blob/feature/Dockerfile

src/main/webapp/search/Search.js
rushtong: Thank you @vvicario. I do understand that the additional styling is necessary in this file - I'm not suggesting that it isn't necessary. In this comment, I'm describing two ways in which a shared css approach is a problematic one. This is also in response to a slack conversation Jaime and I were having about this subject.

src/main/webapp/components/InternationalCohorts.css
vvicario: I agree. We will try to do that in the following changes

src/main/webapp/dataUseLetter/DataUseLetter.js
vvicario: Done!

tadeork (approved):
:+1:

src/main/webapp/components/TopNavigationMenu.css
jaime-bernales-broad: @rushtong Search list styles are now included in the TopNavigationMenu.js and removed from CSS file

LeoForconesi (approved):
:+1:

src/main/webapp/components/TopNavigationMenu.js
```diff
@@ -0,0 +1,294 @@
...
+              span({ className: "sr-only" }, ["Toggle navigation"]),
+              span({ className: "icon-bar" }, []),
+              span({ className: "icon-bar" }, []),
+              span({ className: "icon-bar" }, [])
+            ]),
+            h(Link, { className: "navbar-brand", to: { pathname: Storage.userIsLogged() ? UrlConstants.index : '/' } },
+              [
+                span({}, [
+                  "ORSP Portal ",
+                  span({ isRendered: process.env.NODE_ENV === 'development', className: "label label-danger" }, ["Dev"])
```
rushtong: Note that this still seems like an unresolved question. Where is `process.env.NODE_ENV` being set at the environment level? We do use it in at least one other place, but I don't see where we are setting this value.

rushtong (approved):
Tentatively approving this. If we can find where `process.env.NODE_ENV` is being set, I'm happy with the changes. 👍

src/main/webapp/components/TopNavigationMenu.js
vvicario: Hi Greg. I think this link has the information that we need: 
https://create-react-app.dev/docs/adding-custom-environment-variables

The environment variables are embedded during the build time. These environment variables will be defined on process.env. There is also a special built-in environment variable called NODE_ENV. You can read it from process.env.NODE_ENV. When you run npm start, it is always equal to 'development', when you run npm test it is always equal to 'test', and when you run npm run build to make a production bundle, it is always equal to 'production'. These environment variables can be useful for displaying information conditionally based on where the project is deployed or consuming sensitive data that lives outside of version control.

rushtong: OK, that helps explain where it *could* be used. The application doesn't run under npm, it's a native java app. I'm pretty sure this doesn't exist in our current build environment, so this feature doesn't look like it will work. I'm also fine with moving this to a separate story since the implementation is pretty unrelated to the goal of the PR. In that story, we should address where the other use of it is as well - that is likely broken.

vvicario: Greg. I removed process.env.NODE_ENV. I'm getting this value from another place.

<%@ page import="grails.util.Environment" %>
 env: "${Environment.current.name}"

rushtong: Looks good! 👍

PR #276, opened 2019-09-23
Check non broad user in About and Landing page

vvicario:
## Addresses
No Jira ticket

## Changes
Remove error message from gsp and add it in about page

## Testing
Login with non Broad User, error message should be displayed
---


rushtong (approved):
Thank you for splitting this out. In production now, there are only two dev addresses that are non-Broad (grushton dev at gmail.com and orsp dev at gmail.com) so there are no practical problems introduced with this conversion. I do anticipate that this might be too restrictive in the long run now that the portal site is open to the world.

dags (approved):

PR #277, opened 2019-09-23
[BTRX-792] Project buttons auth

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-792

## Changes
Changed the validation to hide or display buttons on Project review page.

## Testing

---

### DIFF
--- a/src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
+++ b/src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
@@ -26,7 +26,7 @@ const styles = {
   link: {
       textDecoration: "none",
       display: "inline-block",
-      height: "30px",
+      height: "42px",
       lineHeight: "30px",
       color:"#000000",
   }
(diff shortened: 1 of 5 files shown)


src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
```diff
@@ -80,13 +80,13 @@ export const SampleDataCohortsCollapsibleHeader = hh(class SampleDataCohortsColl
...
             button({
-              isRendered: status === STATUS.pending || isEmpty(status) && component.isAdmin,
+              isRendered: component.isAdmin && status === STATUS.pending || isEmpty(status),
               className: 'btn btn-default btn-sm confirmationModal',
               style: styles.pointer.auto,
               onClick: (e) => approveHandler(e, projectKey)
             },['Approve']),
             button({
-              isRendered: status === STATUS.pending || isEmpty(status) && component.isAdmin,
+              isRendered: component.isAdmin && status === STATUS.pending || isEmpty(status),
```
rushtong: I really wish I had caught this earlier. Having three booleans here is very difficult to reason about. From a js repl:
```
true && true || true => true
true && true || false => true
true && false || true => true
false && true || true => true
true && false || false => false
false && true || false => false
false && false || false => false
```
I suggest parentheses or extracting to a function that is easier to read/reason about is a better way to make this check.

rushtong (commented):
Looks good, one request inline about the triple boolean check. If you want to tackle that, no need for a re-review from me.

src/main/webapp/CollapsiblePanel/SampleDataCohortsCollapsibleHeader.js
tadeork: :+1: Yes, totally agree.

PR #278, opened 2019-09-23
BTRX - 790 - Fix legacy institutional resource

LeoForconesi:
## Addresses
[BTRX-790](https://broadinstitute.atlassian.net/browse/BTRX-790): Institutional Sources not working on Legacy Sample/Data Cohorts
## Changes
- Added an extra condition for legacy projects where `institutionalSources` does not exist on the database.

## Testing
- Open a legacy consent group or a consent group that doesn't have any institutional Source stored in the data base, add a new element, the should allow to do this.
- Create a new consent group, its functionality shouldn't have changed

---

### DIFF
--- a/src/main/webapp/components/InstitutionalSource.js
+++ b/src/main/webapp/components/InstitutionalSource.js
@@ -24,12 +24,14 @@ export const InstitutionalSource = hh(class InstitutionalSource extends Componen
     }
   }
 
+  checkEmptySrc(elementPostition) {
+    return (get(this.props.institutionalSources[elementPostition], 'future.name') !== '' && get(this.props.institutionalSources[elementPostition], 'future.country') !== '')
+      || (!isEmpty(this.props.institutionalSources[elementPostition].current.name) && !isEmpty(this.props.institutionalSources[elementPostition].current.country));
+  }
+
   addInstitutionalSources() {
     if (this.props.edit) {
-      if (!isEmpty(this.props.institutionalSources) &&
-         (get(this.props.institutionalSources,'[0].future.name') !== '' && get(this.props.institutionalSources,'[0].future.country') !== '')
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/InstitutionalSource.js
```diff
@@ -25,8 +25,8 @@ export const InstitutionalSource = hh(class InstitutionalSource extends Componen
   addInstitutionalSources() {
     if (this.props.edit) {
       // Only for edit / review
-      if ((this.props.institutionalSources[0].future.name !== '' && this.props.institutionalSources[0].future.country !== '')
-        || this.props.institutionalSources[0].current.name && this.props.institutionalSources[0].current.country ) {
+      if (this.props.institutionalSources[0] && (this.props.institutionalSources[0].future.name !== '' && this.props.institutionalSources[0].future.country !== '')
+        || this.props.institutionalSources[0] && (this.props.institutionalSources[0].current.name && this.props.institutionalSources[0].current.country )) {
```
rushtong: This is a boatload of property checking ... I count 6 in this one line. If we find ourselves needing this, we should be looking for a better solution. Maybe a combination of lodash get and isEmpty would make this easier to read and reason about. 

* https://lodash.com/docs/4.17.15#get
* https://lodash.com/docs/4.17.15#isEmpty

rushtong (commented):
Looks good, one suggestion inline. No need for a re-review from me if you want to tackle that.

tadeork (approved):

src/main/webapp/components/InstitutionalSource.js
LeoForconesi: I've changed this as you suggested. used Get and IsEmpty from lodash.

PR #288, opened 2019-09-27
Delete File on Submission creation

wloforteBroad:
## Addresses
No Jira ticket

---

### DIFF
--- a/src/main/webapp/projectContainer/SubmissionForm.js
+++ b/src/main/webapp/projectContainer/SubmissionForm.js
@@ -241,9 +241,8 @@ const SubmissionForm = hh(class SubmissionForm extends Component {
   };
 
   updateDocuments = () => {
-    const documentsToUpdate = this.state.documents.filter(doc => doc.id !== this.state.fileToRemove.id);
     this.setState(prev => {
-      prev.documents = documentsToUpdate;
+      prev.documents = prev.documents.filter(doc => doc.id !== this.state.fileToRemove.id);
       return prev;
     });
     this.props.hideSpinner();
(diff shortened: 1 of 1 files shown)


LeoForconesi (approved):

src/main/webapp/projectContainer/SubmissionForm.js
```diff
@@ -224,14 +224,28 @@ const SubmissionForm = hh(class SubmissionForm extends Component {
...
-        return prev;
+    if(!isEmpty(this.state.fileToRemove.uuid)) {
+      ProjectMigration.removeSubmissionFile(this.state.params.submissionId, this.state.fileToRemove.uuid).then(prev => {
+        this.updateDocuments();
+      }).catch(error => {
+        console.error(error);
+        this.setState(prev => {
+          prev.errors.serverError = true;
+          return prev;
+        });
```
rushtong, line 235: As a side note, I prefer this style of setting state even if there is no dependency on previous state. Using the same pattern of `setState` is preferable to using many different patterns and leads to less bugs introduced down the road. It's easy to fall into the trap of using the other form ... then someone adds a line that does depend on previous state.

src/main/webapp/projectContainer/SubmissionForm.js
```diff
@@ -224,14 +224,28 @@ const SubmissionForm = hh(class SubmissionForm extends Component {
...
       });
-      this.props.hideSpinner();
+    } else {
+      this.updateDocuments();
+    }
+  };
+
+  updateDocuments = () => {
+    this.setState(prev => {
+      prev.documents = prev.documents.filter(doc => doc.id !== this.state.fileToRemove.id);
```
rushtong, line 245: You can also use lodash's `_.difference` here, but this works.

rushtong (approved):
👍

PR #292, opened 2019-10-02
BTRX - 801 - External DUL ontology service

LeoForconesi:
## Addresses
[BTRX-801](https://broadinstitute.atlassian.net/browse/BTRX-801): Ontology service is not working in External DUL when user is not logged in

## Changes
- Method getMatchingDiseaseOntologies was moved to DataUseLetter Controller. This controller does not depend on Authenticated Controller, element that is interfering on this unauthenticated page.

## Testing
- Create a Data Use Letter Link from a consent group's Document tab. Use that link in a Un-logged session, select from the section 1. Primary Restrictions, option Future research is restricted to (a) specific disease(s):, and then check option "Other", type some disease here and the autocomplete should work as expected.

---


rushtong (approved):
👍 Thanks for the explanation in the PR description, makes sense.

vvicario (approved):

PR #296, opened 2019-10-04
BTRX - 803 - Format text with HTML tags on display

LeoForconesi:
## Addresses
[BTRX-803](https://broadinstitute.atlassian.net/browse/BTRX-803): Format text with HTML tags on display

## Changes
- `InputFieldTextArea` have been replaced by TinyMce editor on **SubmissionForm** in order to create a controlled string with html tags, so later in **Submissions** component we can use `dangerouslySetInnerHTML` without worring on a possible XXS attack coming from any previously mentioned component.
- Adding some missing style classes on SubmissionForm buttons.

## Testing
- You should be able to see a text formatter when adding, or editing a Submission in its tab from a project page.
- After an edition or adding a new one the format should be visible in the submissions table. Also if the text already had a html format (for legacy submissions).

---


rushtong (approved):
👍 Looks right. I had to do the same kind of raw-html display in the gsps.

tadeork (approved):

PR #243, opened 2019-08-20
BTRX 758 ORSP: Show Sample Data Cohort Docs in CG Documents Tab

vvicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-758
## Changes
Consent Collection Link documents are displayed under consent group document  tab
## Testing
Create a new Sample Data Cohort, documents added in creation process should be displayed  under consent group document  tab
---

- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Get a thumb from Belatrix
- [ ] **Submitter**: Get a thumb from @rushtong
- [ ] **Submitter**: Merge to develop using github's "Squash and Merge" feature
- [ ] **Submitter**: Delete branch after merge

PR #303, opened 2019-10-14
BTRX - 819 - Documents are not shown properly

LeoForconesi:
## Addresses
[BTRX-819](https://broadinstitute.atlassian.net/browse/BTRX-819) : Documents are not shown properly

---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
@@ -6,7 +6,6 @@ import groovy.util.logging.Slf4j
 import groovyx.net.http.FromServer
 import groovyx.net.http.HttpBuilder
 import groovyx.net.http.OkHttpEncoders
-import org.apache.commons.collections.CollectionUtils
 import org.apache.commons.lang.StringUtils
 import org.broadinstitute.orsp.config.ConsentConfiguration
 import org.broadinstitute.orsp.consent.ConsentAssociation
@@ -521,11 +520,12 @@ class ConsentService implements Status {
      * @return List of Consent Group that are related to the specified consent collection links
      */
     Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
-        if (keys && !keys.isEmpty()) {
-            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
+        if (!keys?.isEmpty()) {
+            List<Issue> issues = queryService.findAllByProjectKeyInList(keys.keySet().toList(), null)
+            List<StorageDocument> documents = queryService.getAttachmentsForProjects(keys.keySet())
+            def docsByProject = documents.groupBy({d -> d.projectKey})
             issues.each { issue ->
-                // set all documents related to the specified consent group including its consent collection link documents
-                issue.setAttachments(queryService.getAttachmentsForProjects(keys.keySet()))
+                issue.setAttachments(docsByProject.get(issue.projectKey))
                 issue.setStatus(keys.get(issue.projectKey).status)
             }
             issues
--- a/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
@@ -503,7 +503,7 @@ class ConsentService implements Status {
             Issue issue = queryService.findByKey(projectKey)
             Collection<ConsentCollectionLink> collectionLinks = ConsentCollectionLink.findAllByProjectKey(issue.projectKey)
             Map<String, ConsentCollectionLink> collectionLinksMap = collectionLinks?.collectEntries{[it.consentKey, it]}
-            Collection<Issue> consentGroups = findConsentGroupsAssociatedByConsentLink(collectionLinksMap)
+            Collection consentGroups = queryService.findByKeys(collectionLinksMap, projectKey)
             [
                 issue        : issue,
                 consentGroups: consentGroups?.sort { a, b -> a.summary?.toLowerCase() <=> b.summary?.toLowerCase() }
@@ -511,26 +511,6 @@ class ConsentService implements Status {
         } else {
             log.error("Error trying to get Project's Consent groups: Empty projectKey")
             throw new IllegalArgumentException("Error trying to get Project's Consent groups: Empty projectKey")
-        }
-    }
-
-    /**
-     * Find all consent groups associated to the specified consent collection links
-     *
-     * @param keys The consent group key and related consent collection link
-     * @return List of Consent Group that are related to the specified consent collection links
-     */
-    Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
-        if (keys && !keys.isEmpty()) {
-            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
-            issues.each { issue ->
-                // set all documents related to the specified consent group including its consent collection link documents
-                issue.setAttachments(queryService.getAttachmentsForProjects(keys.keySet()))
-                issue.setStatus(keys.get(issue.projectKey).status)
-            }
-            issues
-        } else {
-            Collections.emptyList()
         }
     }
 
--- a/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
@@ -6,6 +6,7 @@ import groovy.util.logging.Slf4j
 import groovyx.net.http.FromServer
 import groovyx.net.http.HttpBuilder
 import groovyx.net.http.OkHttpEncoders
+import org.apache.commons.collections.CollectionUtils
 import org.apache.commons.lang.StringUtils
 import org.broadinstitute.orsp.config.ConsentConfiguration
 import org.broadinstitute.orsp.consent.ConsentAssociation
@@ -502,7 +503,7 @@ class ConsentService implements Status {
             Issue issue = queryService.findByKey(projectKey)
             Collection<ConsentCollectionLink> collectionLinks = ConsentCollectionLink.findAllByProjectKey(issue.projectKey)
             Map<String, ConsentCollectionLink> collectionLinksMap = collectionLinks?.collectEntries{[it.consentKey, it]}
-            Collection consentGroups = queryService.findByKeys(collectionLinksMap, projectKey)
+            Collection<Issue> consentGroups = findConsentGroupsAssociatedByConsentLink(collectionLinksMap)
             [
                 issue        : issue,
                 consentGroups: consentGroups?.sort { a, b -> a.summary?.toLowerCase() <=> b.summary?.toLowerCase() }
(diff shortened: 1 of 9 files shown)


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1277,7 +1277,8 @@ class QueryService implements Status {
                 ' from storage_document d ' +
                 ' where d.id not in (select distinct storage_document_id from submission_document) ' +
                 ' and d.project_key = :projectKey' +
-                ' and d.deleted = 0 '
+                ' and d.deleted = 0 ' +
+                ' and d.consent_collection_link_id is null'
```
rushtong: Conceptually, this looks like the wrong approach. A document, if it is attached to a project, should be returned even if it's also a consent group document. This should only work if the UI is explicit about what it's doing. From an API perspective, this is a very strange restriction to apply at the query level. 

If we need a method that says, give me my documents that are **NOT** in a consent group, then that should be a different method with a much clearer name that describes it as such. Additionally, that query should be called from an API that is clear, i.e., getNonConsentGroupProjectDocuments` or something like that.

rushtong (changes requested):

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -629,16 +629,10 @@ class QueryService implements Status {
...
                 List<StorageDocument> docs = new ArrayList<>()
-                List<ConsentCollectionLink> links = findConsentCollectionLinksByProjectKeyAndConsentKey(projectKey, issue.projectKey)
-                List<StorageDocument> collectionDocuments = findAllDocumentsBySampleCollectionIdList(links?.collect{it.id})
                 List<StorageDocument> projectDocuments = docsByProject.get(issue.projectKey)
-                if (CollectionUtils.isNotEmpty(collectionDocuments)) {
-                    docs.addAll(collectionDocuments)
-                }
                 if (CollectionUtils.isNotEmpty(projectDocuments)) {
                     docs.addAll(projectDocuments)
                 }
```
rushtong: Two things here. 

1. This could be simpler, just set the value since you're no longer summing two collections: `issue.setAttachments(projectDocuments);`
2. Is this still the right thing to do here in `QueryService`? If I am writing an API endpoint, and I want to find all issues by key, I would expect to get *ALL* documents for the issues I want, not *some* of the documents for the issues I want ... then have to know that the *some* doesn't include consent documents, and then have to go retrieve those again. My previous suggestion was to re-think the API you're expecting this query to populate. That's where we can either say, *query for projects without CG docs* OR *query for projects* and post filter docs that don't make sense in this context.

rushtong (changes requested):
Similar feedback to last time.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
verovicario: @rushtong I agree. I moved this logic to Consent Service. I'm getting all documents for the specified key, including link documents if they exist. 
The reason I did that is because we got duplicated documents, we looked for all the files and after that we looked for link documents that we already have in the previous query.

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -513,6 +514,26 @@ class ConsentService implements Status {
...
 
+    /**
+     * Find all consent groups associated to the specified consent collection links
+     *
+     * @param keys The consent group key and related consent collection link
+     * @return List of Consent Group that are related to the specified consent collection links
+     */
+    Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
+        if (keys && !keys.isEmpty()) {
+            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
```
rushtong: Prefer raw queries over GORM here.

rushtong: Isn't this the same thing as `QueryService.findAllByProjectKeyInList` with a null max?

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -513,6 +514,26 @@ class ConsentService implements Status {
...
     }
 
+    /**
+     * Find all consent groups associated to the specified consent collection links
+     *
+     * @param keys The consent group key and related consent collection link
+     * @return List of Consent Group that are related to the specified consent collection links
+     */
+    Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
+        if (keys && !keys.isEmpty()) {
```
rushtong: Could be simpler: `if (!keys?.isEmpty()) ...`

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -513,6 +514,26 @@ class ConsentService implements Status {
...
+            }
+            issues
+        } else {
+            Collections.emptyList()
+        }
+    }
+
     Boolean collectionDocumentsApproved(List<Long> ids){
         List<StorageDocument> documents = queryService.findAllDocumentsBySampleCollectionIdList(ids)
         ArrayList<StorageDocument> pendingDocuments = documents.findAll {
```
rushtong: Technically not related to this PR, but I noticed this while reviewing. This block is trying to coerce a list of strings to a list of `StorageDocuments`:
```
        ArrayList<StorageDocument> pendingDocuments = documents.findAll {
            it.status == DocumentStatus.PENDING.status
        }.fileType
```
Here's one way we could avoid the multiple mutated lists with some the spread operator (*), some chained streaming and null-safe operators:
```
        queryService.
                findAllDocumentsBySampleCollectionIdList(ids)*.
                status?.
                findAll { it == DocumentStatus.PENDING.status}?.
                size() == 0
```
There are probably better ways to answer the question, but this is one.

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -513,6 +514,26 @@ class ConsentService implements Status {
...
+     *
+     * @param keys The consent group key and related consent collection link
+     * @return List of Consent Group that are related to the specified consent collection links
+     */
+    Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
+        if (keys && !keys.isEmpty()) {
+            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
+            issues.each { issue ->
+                // set all documents related to the specified consent group including its consent collection link documents
+                issue.setAttachments(queryService.getAttachmentsForProjects(keys.keySet()))
```
rushtong: I strive to avoid making dao calls inside loops. This leads to all sorts of slowness. It's better to make a new `QueryService` method that queries for all docs on a list of project keys and retrieve them as a map of project key -> list of docs. Then loop over the map and assign the docs to the issue.

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -6,6 +6,7 @@ import groovy.util.logging.Slf4j
 import groovyx.net.http.FromServer
 import groovyx.net.http.HttpBuilder
 import groovyx.net.http.OkHttpEncoders
+import org.apache.commons.collections.CollectionUtils
```
rushtong: unused

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
```diff
@@ -513,6 +514,26 @@ class ConsentService implements Status {
...
+     * @param keys The consent group key and related consent collection link
+     * @return List of Consent Group that are related to the specified consent collection links
+     */
+    Collection<Issue> findConsentGroupsAssociatedByConsentLink(Map<String, ConsentCollectionLink> keys) {
+        if (keys && !keys.isEmpty()) {
+            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
+            issues.each { issue ->
+                // set all documents related to the specified consent group including its consent collection link documents
+                issue.setAttachments(queryService.getAttachmentsForProjects(keys.keySet()))
+                issue.setStatus(keys.get(issue.projectKey).status)
```
rushtong: I'm confused by this ... why are we changing the status of query results? Shouldn't the status be what the status is from the database? As a developer or a user, this would be terribly confusing in the case that these were different. I would see one status in one view and a different status in a different view.

rushtong (changes requested):
With the number of different approaches taken so far, I'm concerned that we don't have a clear implementation plan for this story. Please write up a technical document that explains the problem so we can explore potential solutions. 
FYI @wloforteBroad, @dags

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
verovicario: I agree with you, I didn't want to refactor the existing code, I just only moved the method and I tried only to change documents logic. But it's a good opportunity to improve the existing code

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
verovicario: It was confusing for me too. This method return the information that Project -> Sample / Data Cohort tab needs. Status is consent collection link status not issue. I think it would be helpful to return a DTO with the information that it's needed. 

image_2019_10_17T13_13_04_295Z

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
verovicario: Yes, it's the same, I think this logic exists before we created findAllByProjectKeyInList, I'll improve that

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
verovicario: Sure!

verovicario:
@rushtong I created a document, I hope this  help to clarify it. Thanks!

https://docs.google.com/document/d/14CpJNMl99TIT0I82LjCIzgml2fkWDvXYRXNbO6IBs0M/edit

rushtong:
Here is the original method that needs revision:
```
    Collection<Issue> findByKeys(Map<String, ConsentCollectionLink> keys, String projectKey) {
        if (keys && !keys.isEmpty()) {
            List<Issue> issues = Issue.findAllByProjectKeyInList(keys.keySet().toList())
            List<StorageDocument> documents = getAttachmentsForProjects(keys.keySet())

            def docsByProject = documents.groupBy({d -> d.projectKey})
            issues.each { issue ->
                List<StorageDocument> docs = new ArrayList<>()
                List<ConsentCollectionLink> links = findConsentCollectionLinksByProjectKeyAndConsentKey(projectKey, issue.projectKey)
                List<StorageDocument> collectionDocuments = findAllDocumentsBySampleCollectionIdList(links?.collect{it.id})
                List<StorageDocument> projectDocuments = docsByProject.get(issue.projectKey)
                if (CollectionUtils.isNotEmpty(collectionDocuments)) {
                    docs.addAll(collectionDocuments)
                }
                if (CollectionUtils.isNotEmpty(projectDocuments)) {
                    docs.addAll(projectDocuments)
                }
                issue.setAttachments(docs)
                issue.setStatus(keys.get(issue.projectKey).status)
            }
            issues
        } else {
            Collections.emptyList()
        }
    }
```

My suggestion is to simplify it - it's doing far too much work, adding complexity, and contains at least two bugs ... duplicate documents and mismatched statuses. My suggestion works conceptually, i.e. it only adds one set of docs to each issue and doesn't try to do anything else other than what a query method should do. 
... slightly edited from my first version:
```
    Collection<Issue> findByKeys(Map<String, ConsentCollectionLink> keys) {
        if (!keys.isEmpty()) {
            List<Issue> issues = findAllByProjectKeyInList(keys.keySet().toList(), null)
            List<StorageDocument> documents = getAttachmentsForProjects(keys.keySet())
            def docsByProject = documents.groupBy({it.projectKey})
            issues.each { it.setAttachments(docsByProject.get(it.projectKey)) }
            issues
        } else {
            Collections.emptyList()
        }
    }
```
Looking at the method signature more closely, why do we need the map? Isn't it the case that we just need the list of project ids? And ... if that is the case ... this is just a more robust version of `findAllByProjectKeyInList` since it adds storage docs. This should be renamed to `findAllByProjectKeyInListWithAttachments` or something like that.

rushtong:
A good chunk of the problematic method was introduced in https://github.com/broadinstitute/orsp-pub/pull/243 and I didn't catch it then. My apologies @verovicario - I should have seen where this was going off the tracks.

grails-app/services/org/broadinstitute/orsp/ConsentService.groovy
rushtong: RE:
> Status is consent collection link status not issue

The method in question here is not a UI method, it's an underlying query method. We shouldn't be manipulating the underlying data like this in a query method. If the UI needs to display link status, then the UI should be querying for the links and showing that in the right place.

LeoForconesi:
:+1:

tadeork (approved):
As we see together with @verovicario there was added a DTO to reflect the specific status of the consent collection link, it's OK.

rushtong (approved):

PR #308, opened 2019-10-18
BTRX - 830 - Projects API is sending the incorrect status

LeoForconesi:
## Addresses
[BTRX-830](https://broadinstitute.atlassian.net/browse/BTRX-830):  Projects API is sending the incorrect status

## Changes
- Since an issue object posses two statuses, it was needed a logic to get old status if the new one is 'Legacy', or get the new otherwise. This logic was already done in Issue's model getter named `getApprovalStatus`. So the fix is done using this getter in the Json that the api builds for this endpoint.

## Testing
- Using a Rest Client hit the endpoint -> <Path>/api/projects
- The value for status, should always match with orsp application. 

---


rushtong (approved):
👍

dags (approved):

PR #314, opened 2019-10-25
Implement search by Sample Collection ID

verovicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-823
## Changes
Add a sample collection field in a search page to filter.
## Testing
Go to search page. search by Sample Collection ID
---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -756,6 +756,7 @@ class QueryService implements Status {
                 ' left outer join issue_extra_property p on p.issue_id = i.id ' +
                 ' left outer join user u on p.value = u.user_name ' +
                 ' left outer join funding f on f.project_key = i.project_key ' +
+                ' left outer join consent_collection_link ccl on ccl.consent_key =  i.project_key ' +
                 ' where '
 
         // TODO: Streamline the parameterization here, perhaps with a QueryOptions utility method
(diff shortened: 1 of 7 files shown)


tadeork (approved):
:+1:

LeoForconesi (approved):
Looks good! :+1:

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -756,6 +756,7 @@ class QueryService implements Status {
                 ' left outer join issue_extra_property p on p.issue_id = i.id ' +
                 ' left outer join user u on p.value = u.user_name ' +
                 ' left outer join funding f on f.project_key = i.project_key ' +
+                ' left outer join consent_collection_link ccl on ccl.consent_key =  i.project_key ' +
```
rushtong, line 759: This join looks like a bug to me. In this method, we're looking for Issues, which have a `project_key` column. In the consent collection link table, the join column on issue is also `project_key`. I believe this query should be: 
```
' left outer join consent_collection_link ccl on ccl.project_key = i.project_key ' +
```
In it's current form, this will only find joins when the issue is a consent group type issue.

rushtong: Another possibility is to join on either case. If the issue is a consent group, **AND** you want all CCLs in the results for the consent group, then you can also join on the consent key:
```
`left outer join consent_collection_link ccl on ccl.project_key = i.project_key or ccl.consent_key = i.project_key`
```

PR #318, opened 2019-10-30
[BTRX-826] Number edition submition

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-826

## Changes
When editing a Submission the number can be changed to any other that doesn't already exists.

## Testing
Create a new submission, change the number.
Edit the number of a submission that already exists -> should fail.
Edit the number of a submission to any other -> should pass.

---


grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -84,6 +84,7 @@ class SubmissionController extends AuthenticatedController {
                 submission = Submission.findById(params?.submissionId)
                 submission.comments = dataSubmission[0].comments.value
                 submission.type = dataSubmission[0].type.value
+                submission.number = dataSubmission[0].number.value.toInteger()
```
verovicario: check if value is != null before call toInteger()

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -157,6 +158,24 @@ class SubmissionController extends AuthenticatedController {
...
     }
 
+    def validateSubmissionNumber() {
+        String projectKey = params?.projectKey
+        String type = params?.type
+        String number = params?.number
+
+        if (!projectKey && !type && !number) {
+            response.status = 400
+            render([error: "Unable to find number"] as JSON)
```
verovicario: use handleIllegalArgumentException

src/main/webapp/projectContainer/SubmissionForm.js
```diff
@@ -198,15 +208,43 @@ const SubmissionForm = hh(class SubmissionForm extends Component {
...
+        this.setState(prev => {
+          prev.errors.number = true;
+          prev.errors.numberType = 'This number is already used';
+          return prev;
+        });
+        numberError = false;
+      }
+    }
+
+    if (numberError && comment) {
```
verovicario: if this condition is false, this method doesn't have a return value

src/main/webapp/projectContainer/Submissions.js
```diff
@@ -7,6 +7,7 @@ import { Table } from "../components/Table";
 import { Files } from "../util/ajax";
 import _ from 'lodash';
 import { UrlConstants } from "../util/UrlConstants";
+import { Storage } from "../util/Storage";
```
verovicario: nit: Storage is not used

src/main/webapp/util/Storage.js
```diff
@@ -1,6 +1,7 @@
 const CurrentUser = "CurrentUser";
 const UserIsLogged = "isLogged";
 const URLFrom = "URLFrom";
+const SubmissionNumbers = "SubmissionNumbers";
```
verovicario: It's not used

verovicario (changes requested):

verovicario (approved):

LeoForconesi (approved):

grails-app/controllers/org/broadinstitute/orsp/SubmissionController.groovy
```diff
@@ -157,6 +158,23 @@ class SubmissionController extends AuthenticatedController {
...
+            handleIllegalArgumentException(new IllegalArgumentException("invalid params"))
+        }
+
+        Collection<Integer> numbers = queryService.getSubmissionNumber(projectKey, type, number)
+        response.status = 200
+        if (numbers.size() > 0) {
+            render(['valid': false] as JSON)
+        }
+        render(['valid': true] as JSON)
+    }
```
rushtong, line 176: Post-merge comment here in case we ever return to this code. There is a fine line between applying business logic in different layers. In this case, we could be a lot more flexible in the UI if we just returned the collection of submission numbers (and perhaps a map of `type` -> list of `number`) and let the UI decide if the number is valid or not. Another approach might be to add this info to the model for project - that would give the UI more information in a single API call.

PR #323, opened 2019-11-05
BTRX-836 ORSP: Consent Links should not show deleted Projects or Cohorts

verovicario:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-836
## Changes
Filter queries by deleted rows
## Testing
1) Go to project, sample data cohort tab, add a new sample data cohort. Reject it. Cohort must not be listed.
2) Go to data use restriction report, rejected Consent Groups must not been shown
3) Go to Review Category Report, rejected projects and consent groups must not been shown
4) Go to Quality Assurance Report, rejected projects must not be shown
5) Go to Consent Collection Links Report, rejected projects and consent groups must not been shown
---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -1602,7 +1602,7 @@ class QueryService implements Status {
     }
 
     List<ConsentCollectionLink> findCollectionLinks() {
-        String query = 'select ccl.* from consent_collection_link ccl inner join issue i on i.project_key = ccl.consent_key where ccl.deleted = 0 and i.deleted = 0'
+        String query = 'select * from consent_collection_link where deleted = 0 '
         SQLQuery sqlQuery = getSessionFactory().currentSession.createSQLQuery(query)
         List<ConsentCollectionLink> links = sqlQuery.with {
             addEntity(ConsentCollectionLink)
@@ -1621,9 +1621,13 @@ class QueryService implements Status {
 
     PaginatedResponse findAllCollectionLinks(PaginationParams pagination) {
         // get total rows
-        String totalRowsQuery = 'select count(ccl.id) from consent_collection_link ccl ' +
-                ' left join issue project on project.project_key = ccl.project_key left join issue c on c.project_key = ccl.consent_key ' +
-                ' where ccl.deleted = 0 and project.deleted = 0 and c.deleted = 0 ' + getCollectionLinksWhereClause(pagination)
+        String totalRowsQuery =
+                ' select count(ccl.id) from consent_collection_link ccl ' +
+                ' left join issue project on project.project_key = ccl.project_key ' +
+                ' left join issue c on c.project_key = ccl.consent_key ' +
+                ' where ccl.deleted = 0 ' +
+                ' and project.deleted = 0 ' +
+                ' and c.deleted = 0 ' + getCollectionLinksWhereClause(pagination)
         SQLQuery query = getSessionFactory().currentSession.createSQLQuery(totalRowsQuery)
         if (pagination.searchValue) query.setString('term', pagination.getLikeTerm())
         Long totalRows = query.uniqueResult()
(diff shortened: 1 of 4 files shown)


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1592,15 +1592,17 @@ class QueryService implements Status {
...
+            where = ' where (du.consent_group_key LIKE :term OR du.vault_export_date LIKE :term) and i.deleted = 0 '
+        } else {
+            where = ' where i.deleted = 0 '
         }
         where
     }
 
     List<ConsentCollectionLink> findCollectionLinks() {
-        String query = 'select * from consent_collection_link where deleted = 0 '
+        String query = 'select ccl.* from consent_collection_link ccl inner join issue i on i.project_key = ccl.consent_key where ccl.deleted = 0 and i.deleted = 0'
```
rushtong: The inner join, `inner join issue i on i.project_key = ccl.consent_key` restricts the CCLs to those that are only associated to undeleted consent groups. If a CCL is associated to an undeleted project, that CCL will not be captured in this query.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -1619,12 +1621,16 @@ class QueryService implements Status {
 
     PaginatedResponse findAllCollectionLinks(PaginationParams pagination) {
         // get total rows
-        String totalRowsQuery = 'select count(*) from consent_collection_link where deleted = 0 ' + getCollectionLinksWhereClause(pagination)
+        String totalRowsQuery = 'select count(ccl.id) from consent_collection_link ccl ' +
+                ' left join issue project on project.project_key = ccl.project_key left join issue c on c.project_key = ccl.consent_key ' +
+                ' where ccl.deleted = 0 and project.deleted = 0 and c.deleted = 0 ' + getCollectionLinksWhereClause(pagination)
```
rushtong: Generic feedback on the structure of these queries - it's very hard to read them this way. It would be a lot easier if they were broken down into a more friendly format, something like
```
'select count(ccl.id) from consent_collection_link ccl ' +
' left join issue project on project.project_key = ccl.project_key ' + 
' left join issue c on c.project_key = ccl.consent_key ' +
' where ccl.deleted = 0 ' + 
' and project.deleted = 0 ' + 
' and c.deleted = 0 ' + getCollectionLinksWhereClause(pagination)
```

Also, naming is important in these so a dev can see clearly what the join is referring to, i.e., something like:
```
'select count(ccl.id) from consent_collection_link ccl ' +
' left join issue project on project.project_key = ccl.project_key ' + 
' left join issue consent_group on consent_group.project_key = ccl.consent_key ' +
' where ccl.deleted = 0 ' + 
' and project.deleted = 0 ' + 
' and consent_group.deleted = 0 ' + getCollectionLinksWhereClause(pagination)
```

rushtong (changes requested):
Some structural feedback inline, but I also think there is a potential bug in the `findCollectionLinks()` method.

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
verovicario: Right.
If not we can return all the information. 
Currently for each consent group and project we display a link.
If a consent or project was deleted, we can remove the link to prevent  user from navigation to this pages in Consent Collection Links table. 
What do you think?

rushtong: I am not clear about what you are suggesting.

verovicario: Hi Greg, I mean leave the query as it was and in the UI hide the consent value, if it is deleted, but displaying project.

rushtong: This feedback is about the basic functionality of the method. We have to write our underlying query methods so they return correct information. The query, as is, will leave out some qualifying rows. Whatever we do in the UI is completely irrelevant to what this function does. If I were writing this, it would cover all cases regardless of the intended use. 

Something like this (caveat, untested sql here):
```
String query = 'select ccl.* from consent_collection_link ccl 
    inner join issue i on i.project_key = ccl.consent_key or i.project_key = ccl.project_key
    where ccl.deleted = 0 
    and i.deleted = 0' 
```

verovicario:
@rushtong I worked on your feedback. Please re-review it, when you have some time. Thanks

rushtong (approved):
Thanks for the changes 👍

tadeork (approved):
:smiley:

PR #328, opened 2019-11-11
Fix handle session

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-818

## Changes
Added the redirection to the last page where the user was when re-login for _Add existing CG_ and _New Project_.

## Testing
Go to the Add _Existing Sample/Data Cohort_ page, wait one hour and try to submit the link. It should try to re login the user and then take it back to the previous page.

---

### DIFF
--- a/src/main/webapp/linkWizard/LinkWizard.js
+++ b/src/main/webapp/linkWizard/LinkWizard.js
@@ -4,7 +4,7 @@ import { Wizard } from '../components/Wizard';
 import { SelectSampleConsent } from './SelectSampleConsent';
 import { LinkQuestions } from './LinkQuestions';
 import { ConsentCollectionLink, User } from '../util/ajax';
-import { isEmpty } from '../util/Utils';
+import { handleUnauthorized, isEmpty } from '../util/Utils';
 import '../index.css';
 import * as qs from 'query-string';
 import LoadingWrapper from '../components/LoadingWrapper';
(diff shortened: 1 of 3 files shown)


verovicario (approved):

src/main/webapp/linkWizard/LinkWizard.js
```diff
@@ -242,10 +242,14 @@ const LinkWizard = hh( class LinkWizard extends Component {
...
         this.props.history.push('/project/main?projectKey=' + qs.parse(this.props.location.search).projectKey + '&tab=consent-groups&new', {tab: 'consent-groups'});
       }).catch(error => {
-        console.error(error);
-        this.toggleSubmitError();
-        this.changeSubmitState();
-        this.props.hideSpinner();
+        if (error.response != null && error.response.status === 401) {
+          handleUnauthorized(this.props.history.location);
+        } else {
+          console.error(error);
```
rushtong, line 248: Do we need this?

rushtong (approved):
Minor comment inline, but 👍 if that's addressed.

src/main/webapp/linkWizard/LinkWizard.js
tadeork: We found that here was necessary too to handle the unauthorized session.

PR #333, opened 2019-11-15
[BTRX-837] Security question info link

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-837

## Changes
In the info link page now the questions are displayed. 

## Testing

---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -366,7 +366,7 @@ class QueryService implements Status {
         final String query =
             ' select c.id id, c.consent_key consentKey, c.project_key linkedProjectKey, c.pii pii, c.compliance compliance, c.sharing_type sharingType , c.text_sharing_type textSharingType, ' +
                     ' c.text_compliance textCompliance, c.require_mta requireMta, c.sample_collection_id sampleCollectionId, ' +
-                    ' c.phi phi, c.pii_dt piiDt, c.text_store textStore, c.publicly_available publiclyAvailable, c.store store, c.external_availability externalAvailability ' +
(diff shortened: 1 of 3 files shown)


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -366,6 +366,7 @@ class QueryService implements Status {
         final String query =
             ' select c.id id, c.consent_key consentKey, c.project_key linkedProjectKey, c.pii pii, c.compliance compliance, c.sharing_type sharingType , c.text_sharing_type textSharingType, ' +
                     ' c.text_compliance textCompliance, c.require_mta requireMta, c.sample_collection_id sampleCollectionId, ' +
+                    ' c.phi phi, c.pii_dt piiDt, c.text_store textStore, c.publicly_available publiclyAvailable, c.store store, c.external_availability externalAvailability, c.genomic_data genomicData, ' +
```
rushtong, line 369: One of the things I've noticed is a pattern to query for many specific fields instead of querying for everything. I think it makes more sense to select c.*, etc. instead of making these kinds of small changes over time. This is something that I'm dealing with a lot in the legacy Consent queries and I would recommend moving away from this pattern. If we queried for all fields, then this service method would not need any changes and the changes would be limited to the returned object.

So, not a specific request for this PR, but generic feedback that will make development more natural and flexible over time.

rushtong (approved):
👍

grails-app/services/org/broadinstitute/orsp/QueryService.groovy
tadeork: I agree with you. I just followed the pattern but it can be improved.

verovicario (approved):

dags (approved):

PR #335, opened 2019-11-20
[BTRX-803] Format html tags

tadeork:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-803

## Changes
In some description of activities in Projects still shows some tags.

## Testing
To see this I recommend to got to this project *ORSP-2861*, in the activities description are a lot fo html tags and some unicode characters such as `&nbsp;`

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -771,8 +771,7 @@
     "array-uniq": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/array-uniq/-/array-uniq-1.0.3.tgz",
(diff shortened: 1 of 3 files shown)


rushtong (approved):
While this will technically work for the specific condition, this approach is pretty brittle. It would be better to use a library to sanitize the html in some way. On the groovy side, I used https://jsoup.org/. On the react side, something like https://www.npmjs.com/package/sanitize-html-react could be appropriate. I leave it to the team to decide on pursuing a more robust solution.

rushtong (approved):
Much cleaner, thank you for the changes!

verovicario (approved):

PR #344, opened 2019-11-26
ORPS-4 - Add ability to sort My Task list columns. 

verovicario:
## Addresses
Please link to Jira ticket here
https://broadinstitute.atlassian.net/browse/ORSP-4
## Changes
If issue type is consent group, it will return empty approval status. It is applicable only for projects. This value can be set from Admin Only page. This is necessary to order by status field
## Testing
Go to my task list, each column should be sortable
---

### DIFF
--- a/grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
+++ b/grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
@@ -5,7 +5,6 @@ import grails.rest.Resource
 import groovy.util.logging.Slf4j
 import org.broadinstitute.orsp.AuthenticatedController
 import org.broadinstitute.orsp.Issue
-import org.broadinstitute.orsp.IssueType
 import org.broadinstitute.orsp.SupplementalRole
 import org.broadinstitute.orsp.utils.IssueUtils
 
--- a/grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
+++ b/grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
@@ -23,7 +23,7 @@ class IssueListController extends AuthenticatedController {
                     id               : it.id,
                     projectKey       : it.projectKey,
                     summary          : IssueUtils.escapeQuote(it.summary),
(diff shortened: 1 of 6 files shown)


grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
```diff
@@ -22,7 +23,7 @@ class IssueListController extends AuthenticatedController {
                     id               : it.id,
                     projectKey       : it.projectKey,
                     summary          : IssueUtils.escapeQuote(it.summary),
-                    status           : IssueUtils.escapeQuote(it.getApprovalStatus()),
+                    status           : it.type.equals(IssueType.CONSENT_GROUP.getName()) ? '' : IssueUtils.escapeQuote(it.getApprovalStatus()),
```
rushtong: This is confusing to me to be in a view controller. If Consent Groups shouldn't have a status, then that should be true globally and defined at the `Issue` level, or even better, at the database level.

rushtong (changes requested):

grails-app/controllers/org/broadinstitute/orsp/api/IssueListController.groovy
verovicario: I agree with you. I created a changelog to update approval status value for all consent groups. If you have any other comment, please let me know. Thanks!

verovicario:
@rushtong I worked on your feedback. If you have any other comment, please let me know. Thanks!

grails-app/services/org/broadinstitute/orsp/IssueService.groovy
```diff
@@ -407,7 +407,9 @@ class IssueService implements UserInfo {
         Issue newIssue = issue
         newIssue.setRequestDate(new Date())
         newIssue.setUpdateDate(new Date())
-        newIssue.setApprovalStatus("Pending ORSP Admin Review")
+        if (!type.getName().equals(IssueType.CONSENT_GROUP.getName())) {
+            newIssue.setApprovalStatus("Pending ORSP Admin Review")
```
rushtong: There is an `IssueStatus` for this case: `Pending`

rushtong (approved):
👍 One suggestion inline

PR #345, opened 2019-12-02
ORSP-1 When user session times out, change the message displayed

verovicario:
## Addresses

https://broadinstitute.atlassian.net/browse/ORSP-1

## Changes
Update Error Handler to contemplate if error is Unauthorized, message 
## Testing
When user session times out, message should be "Your session has expired, please sign in again."
---

### DIFF
--- a/src/main/webapp/components/ErrorHandler.js
+++ b/src/main/webapp/components/ErrorHandler.js
@@ -37,7 +37,8 @@ export default class ErrorHandler extends Component {
       unauthorized = true;
     }
     this.setState({ 
-      hasError: true
+      hasError: true,
+      unauthorized: unauthorized
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/ErrorHandler.js
```diff
@@ -16,32 +17,44 @@ export default class ErrorHandler extends Component {
...
+          unauthorized: false
         });
       }
     });
   }
  
   componentDidCatch(error, info) {
     console.error(error);
+    let unauthorized = false;
+    if(!isEmpty(error.message) && error.message.includes("401")) {
```
rushtong: Could be simpler to use [lodash get](https://lodash.com/docs/4.17.15#get). Given that `Utils.isEmpty` is essentially a rewrite of a [`lodash function`](https://lodash.com/docs/4.17.15#isEmpty), we should probably deprecate that and use lodash instead - no point in keeping unnecessary code around.

rushtong (approved):
👍 One minor suggestion inline.

src/main/webapp/components/ErrorHandler.js
verovicario: you are right, I'll change it here and in my #346 PR. Thanks!

PR #348, opened 2019-12-05
ORSP-5 Documents added to Cohort should display under Cohort Documents tab

verovicario:
## Addresses
https://broadinstitute.atlassian.net/browse/ORSP-5
https://broadinstitute.atlassian.net/browse/ORSP-12
## Changes
If a document is related to consent collection link its project key was set to null.
This behavior was changed at some point, now documents have project key. If it's related to consent collection link, it must have  a consent key set as project key.
During this transition some documents were not updated to include it, so they have null project key. In order to fix it, we need to update any occurrence where the id is null.
## Testing
It's needed to have some documents related to consent_collection_link with null project_key in storage_document. 
Before run the script:
Go to  the consent group  => documents tab. Documents whose project key is null in storage document will not display.
Run the script:
Go to  the consent group  => documents tab. Documents whose project_key were null in storage_document will display. 
Storage document table shouldn't have documents with null project key as value


rushtong (approved):
This looks right, but have not tested.

PR #439, opened 2020-04-17
ORSP-40 Refactor Projects API

verovicario:
## Addresses
https://broadinstitute.atlassian.net/browse/ORSP-40

## Changes
Create a new method, it will only be used for this API. 
Term Parameter has been removed. Following Atanas Mihalev comment, they don't send any parameter. His comment is in the story. "No parameters are being sent to projects api".

## Testing
Project API: /api/projects
---

### DIFF
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -918,18 +918,13 @@ class QueryService implements Status {
     }
 
     /**
-     * Find all issues
-     * @return List of Issues
+     * Find all issue summaries
+     * @return List of Issues Summaries
      */
-    List<Issue> findIssues() {
-        final String query = 'SELECT * FROM issue WHERE deleted = 0'
-        final session = sessionFactory.currentSession
-        final SQLQuery sqlQuery = session.createSQLQuery(query)
-        final List<Issue> results = sqlQuery.with {
-            addEntity(Issue)
-            list()
-        }
-        results
+    List findIssueSummaries() {
+        final String query = 'SELECT project_key projectKey, summary, type, approval_status approvalStatus, description, status ' +
+                             'FROM issue WHERE deleted = 0'
+        getSqlConnection().rows(query)
     }
 
     /**
--- a/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
+++ b/grails-app/services/org/broadinstitute/orsp/QueryService.groovy
@@ -917,6 +917,26 @@ class QueryService implements Status {
         (params.size() > 0) ? query + " AND ${q} " : query + q
     }
 
+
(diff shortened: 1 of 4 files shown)


grails-app/services/org/broadinstitute/orsp/QueryService.groovy
```diff
@@ -917,6 +917,21 @@ class QueryService implements Status {
...
+    /**
+     * Find all issues
+     * @return List of Issues
+     */
+    List<Issue> findIssues() {
+        final String query = 'SELECT * FROM issue WHERE deleted = 0'
+        final session = sessionFactory.currentSession
+        final SQLQuery sqlQuery = session.createSQLQuery(query)
+        final List<Issue> results = sqlQuery.with {
+            addEntity(Issue)
```
rushtong: `addEntity` is a pretty slow Hibernate operation. You might want to consider handling the results in a more robust mechanism to convert the query responses to an `Issue` object without using ORM.

grails-app/services/org/broadinstitute/orsp/webservice/ApiService.groovy
```diff
@@ -35,15 +35,8 @@ class ApiService {
...
-    Collection<Map<String, Object>> getProjectSummaries(String term) {
-        QueryOptions options = new QueryOptions()
-        if (term) options.freeText = term
-        options.issueTypeNames.addAll(
-                [IssueType.IRB.name,
-                 IssueType.NHSR.name,
-                 IssueType.NE.name,
-                 IssueType.CONSENT_GROUP.name])
-        filterUnique(queryService.findByQueryOptions(options).collect {
+    Collection<Map<String, Object>> getProjectSummaries() {
```
rushtong: I'm pretty sure this method is eventually called from the swagger page (but can't check it for real, it's broken):
```
  /api/projects:
    get:
      tags:
        - Projects
      summary: Project Summaries
      description: Project Summaries
      parameters:
        - in: query
          name: term
          schema:
            type: string
          required: false
          description: Returns projects matching query term against all project fields
      responses:
        200:
          description: Projects
          schema:
            type: array
            items:
              $ref: '#/definitions/Project'
        500:
          description: Server Error
```
And since this method no longer takes a term, the swagger spec is not correct.

rushtong (approved):
👍 See comments inline. Code looks fine, but I think you're losing some functionality and potentially slowing down the overall performance. Could use some performance testing comparisons to more clearly decide if the approach works better or not.

grails-app/services/org/broadinstitute/orsp/webservice/ApiService.groovy
juaniorbegozo: Hi Greg, finally I've added the term functionality again, so swagger it's working now.

